PR #43, opened 2026-04-14
fix: cleanup drag event listeners on component unmount

javier-delacruz:
## Summary

  Fixes a memory leak reported in #1 (the drag-to-create PR).

  If the component unmounts while the user is dragging, the `mousemove` and `mouseup` event
  listeners attached to `document` were never removed.

  Added `onUnmounted` cleanup in both calendar components:
  - `AppointmentCalendar.vue` (weekly view)
  - `AppointmentDailyView.vue` (daily view)


martinezsalmeron (approved):
Thanks for the fix! Clean and well-targeted solution for the memory leak.

Feel free to contribute more if you're interested — PRs are always welcome.

PR #147, opened 2026-07-31
create new vi.json

nguyenhphu:
## Summary

-

## Modules touched

-

### Always

- [ ] PR title follows Conventional Commits (`feat:`, `fix:`, `docs:`, `refactor:`, `chore:`)
- [ ] `cd backend && ruff check . && ruff format --check .` passes
- [ ] `cd frontend && npm run lint` passes
- [ ] Tests added or updated when behaviour changed
- [ ] No cross-module imports outside `manifest.depends`

### If a new module was added

- [ ] Entry point registered in `backend/pyproject.toml` (`[project.entry-points."dentalpin.modules"]`)
- [ ] `backend/app/modules/<name>/CLAUDE.md` written (purpose, public API, events, permissions, gotchas)
- [ ] `backend/app/modules/<name>/CHANGELOG.md` started
- [ ] Alembic migrations live on the module's own branch (`branch_labels=("<name>",)`)
- [ ] `python backend/scripts/generate_catalogs.py` re-run, `docs/modules-catalog.md` and `docs/events-catalog.md` updated
- [ ] If `removable=True`: uninstall round-trip test added (see `backend/tests/test_uninstall_roundtrip.py`)

- [ ] New event added to `backend/app/core/events/types.py` `EventType`
- [ ] `python backend/scripts/generate_catalogs.py` re-run
- [ ] Payload shape documented in module CLAUDE.md (publishers + consumers)

- [ ] Returned by module `get_permissions()` (no module prefix; registry adds it)
- [ ] Listed in `manifest.role_permissions`
- [ ] If user-facing: added to `frontend/app/config/permissions.ts`
- [ ] Endpoint protected with `require_permission("module.resource.action")`

- [ ] Target module is in `manifest.depends`
- [ ] CI `manifest-consistency` is green

- [ ] ADR added under `docs/adr/NNNN-title.md` (copy `docs/adr/TEMPLATE.md`)

- [ ] Appended to `docs/glossary.md` (ES ↔ EN)

- [ ] Its `backend/app/modules/<name>/CHANGELOG.md` updated under `## Unreleased`

## Test plan

- [ ]

### DIFF
--- a/frontend/app/constants/languages.ts
+++ b/frontend/app/constants/languages.ts
@@ -2,5 +2,6 @@ export const SUPPORTED_LOCALES = [
   'en',
   'es',
   'fr',
-  'pt'
+  'pt',
+  'vi'
 ] as const
--- a/frontend/i18n/locales/vi.json
+++ b/frontend/i18n/locales/vi.json
@@ -1,0 +1,2948 @@
+{
+  "app": {
+    "name": "DentalPin",
+    "tagline": "Dental clinic management"
+  },
+  "demo": {
+    "bannerMessage": "Public demo — do not enter real patient data. The database is reset periodically.",
+    "bannerDismiss": "Dismiss",
+    "credentialsTitle": "Demo credentials",
+    "credentialsNote": "For evaluation only — shown because this is a public demo instance.",
+    "copyEmail": "Copy email",
+    "copyPassword": "Copy password",
+    "copied": "Copied to clipboard"
+  },
+  "nav": {
+    "dashboard": "Home",
+    "patients": "Patients",
+    "appointments": "Schedule",
+    "recalls": "Recalls",
+    "treatmentPlans": "Treatment Plans",
+    "budgets": "Quotes",
+    "invoices": "Invoices",
+    "reports": "Reports",
+    "settings": "Settings",
+    "defaultClinicName": "Clinic",
+    "catalog": "Catalog",
+    "menu": "Menu",
+    "openMenu": "Open menu",
+    "close": "Close",
+    "toggleSidebar": "Toggle sidebar"
+  },
+  "auth": {
+    "login": "Log in",
+    "logout": "Log out",
+    "email": "Email",
+    "password": "Password",
+    "loginButton": "Sign in",
+    "loginSuccess": "Signed in successfully",
+    "invalidCredentials": "Incorrect email or password",
+    "emailRequired": "Enter your email",
+    "passwordRequired": "Enter your password",
+    "emailInvalid": "Invalid email address",
+    "accountInactive": "Your account is inactive. Contact your administrator",
+    "tooManyAttempts": "Too many attempts. Try again in a few minutes",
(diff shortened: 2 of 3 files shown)


nguyenhphu:
ok to merge

martinezsalmeron:
Thanks for taking the time to open this — Vietnamese would be a genuinely welcome addition, and I'd be happy to see it land. There's a fair bit more to it than it looks, though, so let me lay out what's needed.

### The blocker

`frontend/i18n/locales/vi.json` is currently byte-identical to `en.json` — none of the ~2900 strings are translated yet. So the language picker would offer "Tiếng Việt", but selecting it would leave the whole app in English. Worth fixing before anything else.

### The rest of the checklist

Adding a locale touches more than the core JSON file. [PR #125](https://github.com/martinezsalmeron/dentalpin/pull/125) (Portuguese) is the best reference to copy — here's the full set:

1. **Module layer locales.** A `vi.json` in each of the 11 module layers under `backend/app/modules/*/frontend/i18n/locales/`, each one registered in that module's own `frontend/nuxt.config.ts` `i18n.locales` block. `@nuxtjs/i18n` v9 does **not** auto-discover layer locale files. (`notifications` uses the prefixed filename `notifications-vi.json`.)
2. **Nuxt UI's own locale** imported into `UApp` in `frontend/app/app.vue` — otherwise date pickers and built-in component strings stay English.
3. **`backend/templates/email/vi/`** — all 11 templates. Not optional: `EmailService._render_template` tries `{locale}/ → default/ → root`, and there is no `default/`, so a missing template raises rather than falling back.
4. **Communications-language regex** in `backend/app/core/auth/router.py` (currently `pattern="^(es|en|fr|pt)$"`), plus the picker in `notifications/frontend/components/settings/ClinicLanguagePage.vue`.
5. **Catalog-name fallback chains.** Five backend resolvers hardcode `es → en → fr → pt` and need `vi` appended: `billing/service.py`, `catalog/tools.py`, `treatment_plan/service.py`, `clinical_notes/service.py::_resolve_label`, `patient_timeline/seed.py`. Miss these and catalog names silently degrade to the internal code on invoices, treatment plans, the timeline and copilot tools.

Out of scope, by precedent: the user manual stays `{en,es}` and the demo seed data stays `es/en/fr`.

### Smaller things

- PR title should follow Conventional Commits — `feat(i18n): add Vietnamese translation`.
- The PR template checklist is still blank; ticking the relevant boxes helps reviewers.

### How to verify it actually works

A host-side `npm run build` won't prove the module locales merged, because `module_layers/` only exists inside the container. Build there instead and compare the output:

```bash
docker-compose exec frontend npx nuxt build
```

Then compare the `vi-*.mjs` chunk count and total bytes against `fr-*`. Modules absent from `modules.json` legitimately contribute no chunk.

---

No rush on any of this, and no obligation to take the whole thing on — if the scope is more than you were expecting, say so and we can figure out a smaller first step. Glad to answer questions along the way.

Issue #56, opened 2026-04-24
core plugin system: uninstall pipeline hardening


martinezsalmeron:
Follow-up to #47 Phase 1. While debugging an "Aplicar cambios" timeout in `/settings/modules`, three pre-existing bugs in the core plugin system came to light. All affect the uninstall path end-to-end — none were caught because the existing tests stop at state transitions and `test_alembic_roundtrip` runs Alembic via subprocess without the rest of the processor.

Two unrelated fixes shipped as part of the Phase 1 work (they blocked the happy path even being exercised):

- `os.kill(os.getpid(), SIGTERM)` → `os.kill(1, SIGTERM)` so respawn works under `uvicorn --reload` where PID 1 is the supervisor.
- `alembic.command.X(...)` called in-process → `subprocess.run(["alembic", ...])` to avoid `RuntimeError: asyncio.run() cannot be called from a running event loop` when `env.py` runs inside the lifespan loop.

The three still-open issues below live in `backend/app/core/plugins/processor.py` and `backend/app/modules/*/__init__.py`.

## 1. Downgrade target is the module's own revision, not its parent

`processor.py::_remove`:

```python
downgrade_target = record.base_revision or "base"
await self._run_downgrade(downgrade_target)
```

`base_revision` is the head of the module's own branch — i.e. the tip that is currently **applied**. Alembic interprets `downgrade <rev>` as "downgrade to rev, leaving rev applied". Net effect: module tables are never dropped.

Observed in a `patient_timeline` uninstall attempt:
```
Running downgrade ... -> pt_0001    ← pt_0001 left applied → tables survive
```
…followed by the lifespan's `alembic upgrade head` immediately re-applying everything above. DB ends with `state=uninstalled` in `core_module` but the `patient_timeline` table still physically exists.

**Fix**: compute the parent of `base_revision` (walk `script.get_revision(base_revision).down_revision`) and use that as the downgrade target. Handle the "no parent" case (module is the first migration in the chain) by downgrading to `"base"`.

## 2. Fase A linear chain makes most `removable=True` modules non-uninstallable in practice

Current linear chain (verified via `alembic heads`):
```
pat_0001 → ... → med_0001 → ag_0001 → tp_0001 → pc_0001 → pt_0001
           → bil_0001 → notif_0001 → sch_0001 → 0002 → 0003 → ag_0002 → ag_0003 → tp_0002
```

[...]

PR #153, opened 2026-08-04
lamanji changes 

lamanji:
## Summary

-

## Modules touched

-

### Always

- [ ] PR title follows Conventional Commits (`feat:`, `fix:`, `docs:`, `refactor:`, `chore:`)
- [ ] `cd backend && ruff check . && ruff format --check .` passes
- [ ] `cd frontend && npm run lint` passes
- [ ] Tests added or updated when behaviour changed
- [ ] No cross-module imports outside `manifest.depends`

### If a new module was added

- [ ] Entry point registered in `backend/pyproject.toml` (`[project.entry-points."dentalpin.modules"]`)
- [ ] `backend/app/modules/<name>/CLAUDE.md` written (purpose, public API, events, permissions, gotchas)
- [ ] `backend/app/modules/<name>/CHANGELOG.md` started
- [ ] Alembic migrations live on the module's own branch (`branch_labels=("<name>",)`)
- [ ] `python backend/scripts/generate_catalogs.py` re-run, `docs/modules-catalog.md` and `docs/events-catalog.md` updated
- [ ] If `removable=True`: uninstall round-trip test added (see `backend/tests/test_uninstall_roundtrip.py`)

- [ ] New event added to `backend/app/core/events/types.py` `EventType`
- [ ] `python backend/scripts/generate_catalogs.py` re-run
- [ ] Payload shape documented in module CLAUDE.md (publishers + consumers)

- [ ] Returned by module `get_permissions()` (no module prefix; registry adds it)
- [ ] Listed in `manifest.role_permissions`
- [ ] If user-facing: added to `frontend/app/config/permissions.ts`
- [ ] Endpoint protected with `require_permission("module.resource.action")`

- [ ] Target module is in `manifest.depends`
- [ ] CI `manifest-consistency` is green

- [ ] ADR added under `docs/adr/NNNN-title.md` (copy `docs/adr/TEMPLATE.md`)

- [ ] Appended to `docs/glossary.md` (ES ↔ EN)

- [ ] Its `backend/app/modules/<name>/CHANGELOG.md` updated under `## Unreleased`

## Test plan

- [ ]


martinezsalmeron (changes requested):
Thanks for the huge amount of work here — 35 commits over three weeks is a real investment, and a lot of it (payroll, the supplier/procurement suite, inventory overhaul, medications catalog, staff activity journal) is stuff this project genuinely needs.

That said, I can't review or merge this as one PR. It currently bundles roughly eight independent features — payroll, procurement (suppliers/supplier_items/purchase_orders/inventory_reorder/supplier_ratings), the inventory core rewrite + auto-deduction, treatment-consumables linkage, medications catalog, staff activity journal, an SMS gateway, plus contacts/tasks/documents/expenses — with unrelated tooling changes (Node 20→22, an `@nuxtjs/i18n` major-version bump, a rewritten `frontend/Dockerfile.prod`, a new Semgrep workflow) on top, across 327 files with the PR description left blank. There's no way to give this a real line-by-line review or reason about what's safe to merge in that shape.

A few concrete things I did spot that need fixing regardless of how it's split:

- **Alembic branch violation (this is the uninstall-safety rule from CLAUDE.md, not a style nit):** `backend/app/modules/treatment_consumables/migrations/versions/tcl_0001_create_treatment_consumables.py:18` and `backend/app/modules/staff_activity/migrations/versions/sact_0001_create_staff_activity_logs.py:18` both set `down_revision = "med_0002"` — that revision actually belongs to the **`media`** module (see `backend/app/modules/media/migrations/versions/med_0002_...py:29`), not `medications`, and neither module declares `media` in `manifest.depends`. Both branches are unknowingly threaded onto media's chain. The payroll migration's own docstring in `backend/app/modules/payroll/migrations/versions/payr_0001_create_payroll_tables.py` says this "had to be manually fixed after the fact" for these two files, but the fix isn't actually in the diff — they're still wrong at HEAD.
- **Undeclared cross-module reach:** `backend/app/modules/inventory_consumption/handlers.py` runs a raw `SELECT catalog_item_id FROM treatments WHERE id = :tid` against a table owned by `odontogram`, which isn't in its `manifest.depends` (`["inventory", "treatment_consumables"]`). If that's intentional to avoid a fourth declared dependency, it still needs `odontogram` added to `depends` per the hard rule — the raw SQL doesn't exempt it.
- **Debug code left in the build config:** `frontend/nuxt.config.ts` (~lines 29-63) writes to `/tmp/debug-layers.json` / `/tmp/debug-resolved-layers.json` on every build via a plugin explicitly commented `// TEMPORARY DEBUG MODULE ... Remove this whole entry once the payroll layer-resolution issue is diagnosed.` — looks like it was never removed.
- **Placeholder migration:** `backend/app/modules/medications/migrations/versions/meds_0002.py` is a literal no-op ("empty migration – placeholder for missing meds_0002").
- Most of the 20 new module directories are missing `CLAUDE.md`/`CHANGELOG.md` and there's no `docs/technical/<name>/` for any of them (the checklist in the repo's CLAUDE.md requires both for new modules) — `lab_orders` is the one that does this right and is a good template to copy from. `notifications`, `patients_clinical`, and `copilot` get real migrations/code changes but no `CHANGELOG.md` entry.
- No test files anywhere in the diff, including for payroll (financial data) and the procurement chain.
- `ruff check` currently reports 38 errors on the new modules (mostly auto-fixable import sorting, but `backend/app/modules/staff_activity/__init__.py:23-27` has mixed tabs/spaces that will fail `ruff format --check`).

To be clear on what's good: the `manifest.depends` declarations you did add are mostly accurate and well thought out (`purchase_orders` → `contacts`/`inventory`/`suppliers`, `inventory_reorder` → all four of its real deps), and `clinic_id` filtering looks consistent in `payroll/service.py` and `purchase_orders/service.py` where I checked.

Could you split this into separate PRs by feature area — at minimum payroll, procurement+inventory, the contacts/tasks/documents/expenses infra, and the unrelated Node/i18n/Docker/CI tooling changes as their own PR? Smaller diffs will get a real review a lot faster than this one can. Happy to help figure out a sane split if useful, and thanks again for the effort that went into this.

---

martinezsalmeron (changes requested):
Thanks for pushing on this, and for merging upstream/main repeatedly to keep the branch alive — that's real work. I took the branch, built it, migrated a database, logged in and drove the new screens in a browser, and reviewed all 343 files. Below is what I found. It's long because the PR is large, and because some of it is serious enough that I want to give you exact reproductions rather than opinions.

**Bottom line: I can't merge this, and the reason is no longer just "it's too big to review".** There are four reproducible security bugs and one data-corruption bug that I confirmed by running them, plus the migration and module-isolation issues from my last review that are still unfixed. My split request still stands, but please read the security section first — several of these bugs exist in the code you're already running in your clinic today.

---

## 1. Confirmed security bugs (I reproduced each of these on this branch)

I created a second clinic in the DB ("QA Tenant B") with one patient, then acted as `admin@demo.clinic`, who belongs to clinic A only.

**1.1 — Cross-tenant patient PII disclosure via lab orders.** `POST /api/v1/lab_orders/` never checks that `patient_id` belongs to the caller's clinic, and `LabOrderService._enrich` (`backend/app/modules/lab_orders/service.py:69-80`) looks up patient and contact names with no `clinic_id` filter. Result:

```
POST /api/v1/lab_orders/  {"patient_id": "<clinic B patient>", "lab_contact_id": "<clinic A lab>", ...}
→ 200 {"clinic_id":"...a11", "patient_name":"SECRETO PacienteClinicaB", ...}
```

Clinic A just read a clinic B patient's full name. With UUID enumeration this is a working PII oracle across every tenant on the instance.

**1.2 — Same hole in document generation, and it renders into a PDF.** `DocumentService._load_patient` (`backend/app/modules/documents/service.py:87-91`) uses `db.get(Patient, patient_id)` with no clinic filter — literally the pattern CLAUDE.md flags as "WRONG — security vulnerability". I generated a prescription from clinic A for the clinic B patient and downloaded the PDF (HTTP 200, 8.6 KB). The stored HTML contains:

```html
<div class="field-row"><span class="field-label">Patient:</span> SECRETO PacienteClinicaB</div>
```

**1.3 — Cross-tenant *write* on clinical reference data.** Every PUT/DELETE in `medical_reference/router.py` fetches by id only (`_get_or_404` → `service.get`, no clinic filter). I renamed clinic B's allergy record from clinic A:

```
PUT /api/v1/medical_reference/allergies/<clinic B allergy id>  {"name":"HACKEADO-POR-CLINICA-A"}
→ 200, and the row in clinic B is now renamed.
```

An allergy list is clinical safety data. Silent cross-tenant edits to it are the worst version of this bug class.

**1.4 — HTML injection into generated documents (and SSRF).** `documents/service.py:66-68` builds the document with `jinja2.Template(...)`, which has autoescape **off** by default. I sent a drug name of `XSSTEST<b>BOLD</b>` and it landed unescaped in the stored HTML and in the PDF. WeasyPrint fetches embedded resources, so `` or `url(file:///etc/passwd)` in any free-text field becomes a server-side request — on generation *and* on every PDF re-render. `letterhead.logo_url` is an unvalidated string that gets fetched server-side too. Fix is `autoescape=True` plus a restricted `url_fetcher`. (`purchase_orders/pdf.py` has the same unescaped interpolation.)

Related: `tasks/service.py:56-62` looks up users by id with no clinic-membership check, so `assigned_to` leaks foreign users' `full_name` the same way.

## 2. Confirmed data corruption

**2.1 — Inventory stock silently loses writes under concurrency.** `InventoryService.record_movement` (`inventory/service.py:110-158`) does read-modify-write with no row lock and no atomic UPDATE. Item at stock 10, five concurrent `-3` movements:

```
5 requests → 200 200 200 200 200
inventory_items.quantity_on_hand = 7.00
inventory_movements: 5 rows, every one recording quantity_after = 7.00
```

Fifteen units were consumed, three were deducted, and the audit trail records the same "after" value five times. Every module funnels stock changes through this method, including the new auto-deduction handler, so this is the one bug I'd fix even if nothing else from this PR ever merges. Needs `SELECT ... FOR UPDATE` or `UPDATE ... SET quantity = quantity + :delta`.

**2.2 — Auto-deduction has no idempotency.** `inventory_consumption/handlers.py:61-74` writes a `used` movement keyed `treatment:{id}` without checking whether one exists, and there's no compensating handler for un-completing a treatment (`treatment_plan` has a `reopen` path). Replay or reopen→re-complete deducts twice. A failed deduction is swallowed with a log line (`handlers.py:75-87`), so stock drifts with no user-visible signal.

**2.3 — Expenses date filter is a 500, not a filter.** `expenses/router.py:30-31` types `date_from`/`date_to` as `str`, and Postgres rejects `date >= character varying`:

```
GET /api/v1/expenses/?date_from=2026-01-01 → 500
"operator does not exist: date >= character varying"
```

Worth noting the raw SQL and the bound parameters come back in the API response body — PR #152 made us surface server messages to users, so DB internals now reach the browser. Type these as `date` and we get a 422 instead.

**2.4 — Purchase order receiving is not atomic.** `receiving_service.py:99-118` calls `InventoryService.record_movement`, which commits on every call (`inventory/service.py:157`). A bad line partway through the loop leaves stock incremented and `quantity_received` bumped with no receipt row and events already published. Also `service.py:53-62` generates PO numbers as `count(*)+1`, so deleting a draft and creating a new PO collides on `uq_purchase_order_clinic_number` and 500s.

## 3. The screens don't actually load

I built the branch with your own `frontend/Dockerfile`, migrated, and logged in. Six of the new pages 404 at runtime because their layers never resolve:

| Route | Result |
|---|---|
| `/documents`, `/contacts`, `/inventory`, `/expenses`, `/payroll`, `/lab-orders` | load |
| `/treatment-consumables`, `/staff-activity`, `/purchase-orders`, `/reorder-suggestions` | **404** |
| `/medications` | **500** — `A <SelectItem /> must have a value prop that is not an empty string` |

The sidebar links to all of them, so a user clicking "Purchase orders" gets a 404 page. Nuxt resolves 29 layers out of the 38 listed in `modules.json`; your own `buildlog.txt` (committed to the PR) shows the `WARN Cannot extend config from /module_layers/...` lines, and `nuxt.config.ts` still carries the debug plugin you added to chase exactly this — self-labelled *"Remove this whole entry once the payroll layer-resolution issue is diagnosed."* It was never diagnosed. **This is the single most useful thing you could fix next**, because until layers resolve deterministically, half of what you built is unreachable and I can't evaluate it.

## 4. Migrations — still broken, and now worse

From my last review, unfixed at HEAD: `treatment_consumables/tcl_0001` and `staff_activity/sact_0001` both set `down_revision = "med_0002"`, which belongs to **media**, not medications. Neither declares `media` in `depends`.

New in this pass — the whole procurement suite is a single chain threaded through four modules that are all marked `removable=True`:

```
inv_0002 (inventory) → supp_0001 (suppliers) → si_0001 (supplier_items) → po_0001 → po_0002 (purchase_orders) → sr_0001 (supplier_ratings)
```

Plus `documents/doc_0001` hangs off `pat_0003` (patients) and `medications/meds_0001` off core `0005`. Uninstalling any link orphans everything downstream — the exact failure mode issue #56 exists to prevent. `medications/meds_0002.py` is a literal no-op placeholder nothing references; delete it.

Practical consequence: `alembic upgrade heads` fails from a clean database. I had to run branches by hand in dependency order to get a working instance. That alone blocks any deploy.

## 5. Module isolation

Imports crossing boundaries with no `depends` declaration:

- `documents` declares `depends: []` but imports `media.service` and `patients.models`, and has a FK to `patients.id` (`doc_0001:45`)
- `purchase_orders` imports `notifications.gateway` (`service.py:27`) — and that path is dead anyway, since `enqueue("purchase_order_sent")` needs a `NotificationTemplate` with that key and nothing anywhere creates one, so "send to supplier" silently emails nobody
- `inventory_reorder` imports `contacts.models` (`service.py:80`) and fetches by PK with no clinic filter
- `medical_reference` imports `patients.service` with only `patients_clinical` declared
- `inventory_consumption` still does `SELECT catalog_item_id FROM treatments` (odontogram's table) in raw SQL with no clinic filter — the docstring says this is to dodge declaring a fourth dependency, which is the rule, not a workaround for it

Dependency direction is also inverted in the frontend: `contacts/pages/contacts/index.vue` imports from the `suppliers` and `supplier_ratings` layers, and `inventory/pages/inventory/index.vue` imports `SupplierItemsModal` from `supplier_items`. Core pages now break if an optional module isn't installed. The slot registry exists for this (`patient_admin/slots.client.ts` uses it correctly — copy that).

## 6. One change that breaks existing behaviour

`frontend/app/config/permissions.ts:68-71` repoints the existing `documents` key from `media.documents.read/write` to `documents.read/write`. That key is already used by `media/components/{DocumentGallery,PhotoGallery,DocumentCard}.vue` and `patients/components/patient/AdministrationTab.vue`, whose backends still enforce `media.documents.*`. Non-admin users now get hidden buttons or visible buttons that 403. The new module needs its own key.

## 7. Things that shouldn't be in the PR at all

`five-fixes-v2.zip`, the unzipped `five-fixes/` copy of it, `FIVE_FIXES_INSTALL_GUIDE.md`, `buildlog.txt`, `tier1_tree.txt` (a UTF-16 Windows listing of `C:\Users\Dr.Amine\Desktop\...` — your machine paths), `.kilo/kilo.jsonc`, and a duplicate `scripts/seed_medications.py`.

Two that need action beyond deletion:

- **`.env.example` is deleted.** That's the onboarding template — the ADR-0006 `BUDGET_PUBLIC_SECRET_KEY` comment, `ALLOWED_ORIGINS`, everything. Please restore it.
- **`backend/data/suppliers_cleaned.csv` contains real third-party PII** — a named contact ("Anis Jallouli") and Tunisian phone numbers, committed to a public repo. Please remove it from the branch. If it's ever been pushed publicly, treat those numbers as disclosed.

`docker-compose.yml` also carries your local setup (frontend moved to 8080, `./frontend:/app` bind mount commented out, `npm run dev` commented out). That deletes the documented hot-reload dev workflow for everyone else.

## 8. Localization mismatch worth discussing openly

A lot of this is built for a Tunisian practice, not a generic dental product: `backend/scripts/seed_data/uploaded_lists_2026_07.json` is the CNAM APCI disease list in French, `supplier_items/models.py:9` says *"Prices are TND only — confirmed with the clinic owner"*, `purchase_orders/pdf.py:108` defaults currency to TND, and the medication seed is a French/Tunisian formulary. DentalPin defaults to EUR/Spain (VeriFactu). This isn't wrong to want — it's a real gap — but it should land as configuration and locale packs, not as hardcoded defaults and seeds. Happy to design that with you; it's a genuinely valuable contribution if we do it properly.

Also on the clinical side: the interaction/contraindication checker only flags pairs a clinic manually typed in, and only for entries with a `reference_id` — legacy free-text entries are skipped silently, so an empty warnings panel is indistinguishable from "checked, no interactions". If we ship this it needs an explicit "advisory only, unlinked entries not checked" disclaimer. Separately, `medications.medications` and `medical_reference_medication` are two unlinked drug catalogs — the 56 seeded drugs never feed the interaction engine.

Related, and this one will annoy your dentists: `medical_reference` grants write to admin only, but the new `ReferenceSearchInput` creates entries on type. A dentist typing a new allergy gets a 403 and the code deliberately leaves the field as-is, so **the allergy simply cannot be recorded**.

## 9. Checklist items (mechanical, but CI enforces them)

- `ruff check backend/app/modules/` → 38 errors; `ruff format --check` → 46 files would be reformatted (`staff_activity/__init__.py` has mixed tabs and spaces)
- `npm run lint` → 444 errors, 545 warnings
- **Zero `docs/` changes in 343 files.** No `docs/technical/<name>/` for any of the 20 new modules; no user-manual pages for ~12 new screens; catalogs not regenerated. 12 of 19 new modules have no `CLAUDE.md`/`CHANGELOG.md`
- **Zero tests**, including payroll (money) and the procurement chain. Every new module is `removable=True` with no uninstall round-trip test
- Most new modules are `auto_install=True`; project policy is `False` with manual activation from the admin UI
- POSTs return 200 instead of 201, DELETEs 200 instead of 204, throughout
- `lab_order.status_changed` is published and consumed as a raw string, absent from `EventType` — and `tasks/CLAUDE.md` says "Events: None" while the code subscribes to it
- Payroll does money math in `float` (`service.py:130,185`), has no state machine (a paid period can be re-processed, paid entries edited), and doesn't verify `user_id` is a member of the clinic
- `staff_activity` writes a row synchronously with its own session and commit on *every* event platform-wide, storing the entire raw payload — patient names and phones included — with no retention policy

---

## What I'd like to do next

I'm not asking you to fix all of this before anything lands. Here's what I think is the fastest path to getting your work into main:

1. **Fix the five security/corruption bugs first, as one small PR** (§1.1–1.4, §2.1). They're each a few lines: add `clinic_id` filters, turn on Jinja autoescape, lock the inventory row. This is worth doing this week regardless of the rest, because you're running this code in a live clinic.
2. **Then the layer-resolution bug** (§3), as its own PR. Until that's fixed, four of your modules don't exist from the user's point of view.
3. **Then split the features** as I asked last time, roughly: (a) contacts/tasks/documents/expenses, (b) inventory + procurement, (c) medications/medical_reference/patient_admin, (d) payroll + staff_activity, (e) the Node 22 / i18n / Docker / CI tooling. Each with its own migration branch off core `0001`, a filled-in PR description, and at minimum a happy-path test.
4. **Drop the junk files, restore `.env.example`, remove the supplier CSV, and revert docker-compose.** That's a five-minute commit and it makes everything else easier to read.

If it helps, I'm happy to do (4) myself and push it to your branch, and to pair on the layer-resolution bug — that one is genuinely non-obvious and it's our build system's fault as much as yours.

To be clear about what's good here, because a lot is: the lab_orders→tasks bridge uses the event bus correctly instead of importing across modules, the SMS credentials are Fernet-encrypted with write-only fields and `has_api_key` in responses, payroll's RBAC is genuinely tight (admin-only, bank details encrypted, nothing exposed to the copilot), the sidebar super-groups degrade gracefully for unmapped modules and preserve permission filtering, en/es/fr key parity is complete across every layer you added, and your `pyproject.toml` entry-point backfill fixed gaps that were stale on main. The `manifest.depends` declarations you did write are mostly accurate. This is someone who understood the architecture and built a lot of real product on top of it.

Thanks again, and sorry for the length of this. Ping me on any of it.

martinezsalmeron:
Update: I see this got split into #154–#160 along the lines I asked for (payroll, procurement, patient-care, lab-orders, infra, UI, and one that turned out to be just a `modules.json` edit). I've reviewed each individually there rather than re-reviewing the combined diff here. Short version: several of the confirmed bugs from my last review here (the lab_orders/medical_reference/documents cross-tenant lookups, the inventory concurrency bug, the staff_activity migration branch issue) are still unresolved in the splits, plus a few new issues from how the split was done (some PRs reference modules/events that only exist in their sibling PRs, so they're not independently mergeable yet).

Leaving this open for now since it's the original umbrella — happy to close it once the split PRs are sorted, whichever way you'd prefer to handle it.

---

PR #155, opened 2026-08-05
chore(infra): Update Docker, environment configuration, Alembic, and event types

lamanji:
Updates core infrastructure and environment configuration files:

Updates docker-compose.yml and frontend/Dockerfile.

Configures backend/alembic.ini and backend/pyproject.toml.

Updates event definitions in backend/app/core/events/types.py.

Adds initial Copilot module setup in backend/app/modules/copilot/.


martinezsalmeron (changes requested):
Thanks for splitting this out. Titled "Docker, environment configuration, Alembic, and event types" but the actual diff is 106 files / 6219 insertions and is dominated by six new feature modules (`contacts`, `documents`, `expenses`, `recall_reminders`, `sms_gateway`, `tasks`) with full backend + frontend + migrations — the infra changes are a small part of it. That mismatch makes this hard to review as "infra."

Things that need fixing regardless of how it's retitled/re-split:

- **Security bug carried over from #153, still present:** `backend/app/modules/documents/service.py:67` builds documents with `jinja2.Template(...)`, which has autoescape off by default — unescaped free text lands in generated HTML/PDFs, and since WeasyPrint fetches embedded resources, that's an SSRF vector on every render (fix: `autoescape=True` + a restricted `url_fetcher`). And `service.py:88` does `await db.get(Patient, patient_id)` with no `clinic_id` filter — a clinic can generate a document for another clinic's patient and get their name back. Both were flagged in the #153 review and are unchanged here.
- **Migration threads through another module's branch:** `backend/app/modules/documents/migrations/versions/doc_0001_documents_initial.py:19` sets `down_revision = "pat_0003"`, chaining off `patients`' own revision instead of rooting on `"0001"` — the file's own docstring admits the real head still needs to be filled in. This is the uninstall-safety rule from CLAUDE.md (issue #56).
- **Undeclared cross-module dependency:** `documents/__init__.py:29` declares `depends: []`, but `service.py:14-15` imports `app.modules.media.service` and `app.modules.patients.models`, and there's an FK to `patients.id`. Needs `depends: ["patients", "media"]` (or the equivalent) declared.
- **Wrong CLAUDE.md:** `backend/app/modules/documents/CLAUDE.md` is a verbatim copy of the `lab_orders` module's doc — wrong module name, wrong API, wrong everything. Looks like a copy-paste that wasn't updated.
- **Phantom event types:** `backend/app/core/events/types.py` adds `INVENTORY_*`, `PURCHASE_ORDER_*`, `PAYROLL_*` events for modules that don't exist in this PR's tree (they're in #158/#159). Same for `backend/pyproject.toml`/`backend/alembic.ini`, which register ~14 modules not present here. Harmless at runtime (loader catches `ImportError`) but means this branch can't stand alone — merging it before #155's sibling PRs land will reference modules that aren't there yet.
- **Missing docs:** none of the six new modules got `CLAUDE.md`/`CHANGELOG.md`, or `docs/technical/<name>/`.
- **Frontend permission not wired:** `contacts/frontend/pages/contacts/index.vue:14` references `PERMISSIONS.contacts.read`, which isn't defined in `frontend/app/config/permissions.ts` in this diff — undefined at runtime.

Given the phantom references, I don't think these seven PRs can be merged independently of each other in their current form — worth sequencing them or accepting they land together. Happy to help figure out a merge order if useful.

---

PR #156, opened 2026-08-05
feat(lab-orders): Add lab orders module, prosthodontic fields, and dashboard widget

lamanji:
Implements end-to-end lab order tracking:

Adds backend/app/modules/lab_orders/ complete with models, schemas, service layer, and router.

Includes migrations labo_0001_initial.py and labo_0002_add_prosthodontic_fields.py.

Adds frontend components, pages (/lab-orders), i18n locales (EN, ES, FR), and LabOrdersDashboardWidget.vue.

### DIFF
--- a/backend/app/modules/lab_orders/CHANGELOG.md
+++ b/backend/app/modules/lab_orders/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog — lab_orders module
+
+## Unreleased
+
+- Initial version: lab order CRUD, status lifecycle (sent → in_progress →
+  ready → received / cancelled), patient + lab contact linkage with
+  enriched display names, EN/ES/FR translations.
--- a/backend/app/modules/lab_orders/CLAUDE.md
+++ b/backend/app/modules/lab_orders/CLAUDE.md
@@ -1,0 +1,68 @@
+# Lab orders module
+
+Tracks work sent to external labs for a specific patient (crown, bridge,
+denture, implant, etc.), with a status lifecycle from "sent" through to
+"received". Custom clinic module.
+
+## Public API
+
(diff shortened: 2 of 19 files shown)


martinezsalmeron (changes requested):
Thanks for splitting this out — the module structure here is genuinely good: `backend/app/modules/lab_orders/CLAUDE.md` and `CHANGELOG.md` are both present and well filled in, `manifest.depends = ["patients", "contacts"]` matches what the code actually imports, and the migration chain is clean (`labo_0001_initial` roots on core `"0001"` with `branch_labels=("lab_orders",)`, `labo_0002` chains off it correctly with no separate branch label). This is the module the #153 review pointed to as the template to copy for docs — still true here.

One thing carried over unfixed from the #153 review though, and it needs to be fixed before this merges: **cross-tenant PII disclosure.** `LabOrderService._enrich` (`backend/app/modules/lab_orders/service.py:68` and `:73`) resolves patient and contact names with `select(Patient).where(Patient.id.in_(patient_ids))` / `select(Contact).where(Contact.id.in_(contact_ids))` — neither filters by `clinic_id`. This method backs every response path (`get_order_response`, `list_order_responses`), so a lab order created with a `patient_id` from another clinic returns that clinic's patient name in the response.

It compounds with a second gap: `create_order` (`service.py:111-130`) validates `lab_contact_id` via `_assert_contact_exists` but has no equivalent check on `patient_id` — so a clinic-A caller can `POST` a clinic-B `patient_id` and get clinic B's patient name back in the 200 response. With UUID enumeration that's a working cross-tenant PII oracle.

Fix is straightforward: add `clinic_id` to both `_enrich` queries, and add a `_assert_patient_exists`-style check (matching the existing contact check) in `create_order` before the record is built.

---

PR #157, opened 2026-08-05
feat(patient-care): Add patient admin, clinical, medical references, and medications

lamanji:
Comprehensive update to patient care and clinical record-keeping:

Patient Admin & Clinical: Adds patient administration module and clinical record reference links (pc_0002, pc_0003).

Medical Reference: Implements medical reference lookup, surgery references, and drug interaction handling.

Medications: Integrates medication search, management UI, and settings pages.

Seeding: Includes seed scripts and datasets for surgeries, medications, and uploaded clinical lists.


martinezsalmeron (changes requested):
Thanks for splitting this out. Good news first: the `meds_0002` no-op placeholder migration from #153 is gone in this split, and migration branch discipline is clean across `medical_reference`, `medications`, and `patient_admin` (each roots on its own core revision with its own `branch_labels`, no threading through another module's chain).

What still needs work before this can merge:

- **Cross-tenant write on clinical reference data, still present — and wider than originally scoped.** `MedicalReferenceService.get` (`backend/app/modules/medical_reference/service.py:66-67`) does `select(model).where(model.id == item_id)` with no `clinic_id` filter. It backs `_get_or_404` (`router.py:47-50`), which every write endpoint goes through: `update_allergy`/`deactivate_allergy` (`router.py:88,102`), `update_medication` (`:148`), and the surgery/disease equivalents. Same pattern independently in `get_interaction` (`service.py:161-162`) and `get_contraindication` (`service.py:233-236`). So it's not just allergies — any clinic can rename or deactivate another clinic's allergies, medications, surgeries, diseases, interactions, or contraindications by guessing an id. This is clinical safety data; the fix is a `clinic_id` filter on each of those lookups.
- **Dentists can't actually record a new allergy.** `role_permissions` in `medical_reference/__init__.py` grants write to admin only, but `ReferenceSearchInput.vue` calls `create()` on type (`useMedicalReference.ts:65-70`), which 403s for a dentist, and the component's own comment says it deliberately leaves the field as-is rather than falling back to free text. So a dentist typing a new allergy silently can't record it at all — worth either widening the write permission to clinical roles or giving the frontend a fallback path.
- **Undeclared import:** `medical_reference/router.py:18` imports `app.modules.patients.service.PatientService`, but `manifest.depends` only lists `patients_clinical`. Even if `patients_clinical` depends on `patients` transitively, CLAUDE.md's rule is about direct imports — please add `patients` to `depends` explicitly.
- **Missing docs:** `medical_reference`, `medications`, and `patient_admin` are all new modules and none has `CLAUDE.md` or `CHANGELOG.md`.
- Minor lint: unsorted imports in `medical_reference/service.py:9` and `patient_admin/service.py:7`, plus two `str + Enum` → `StrEnum` suggestions in `medications/models.py`.

Note the `documents` module (the Jinja-autoescape/SSRF issue from the #153 review) isn't part of this diff — it's in #155, and I've left the same note there.

---

PR #158, opened 2026-08-05
feat(payroll): Add staff activity and payroll management module

lamanji:
Integrates payroll tracking and staff activity monitoring:

Introduces backend service endpoints and models for payroll calculations and staff logging.

Connects frontend UI components for staff activity overview.

### DIFF
--- a/backend/app/modules/payroll/CHANGELOG.md
+++ b/backend/app/modules/payroll/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog — lab_orders module
+
+## Unreleased
+
+- Initial version: lab order CRUD, status lifecycle (sent → in_progress →
+  ready → received / cancelled), patient + lab contact linkage with
+  enriched display names, EN/ES/FR translations.
--- a/backend/app/modules/payroll/CLAUDE.md
+++ b/backend/app/modules/payroll/CLAUDE.md
@@ -1,0 +1,68 @@
+# Lab orders module
+
+Tracks work sent to external labs for a specific patient (crown, bridge,
+denture, implant, etc.), with a status lifecycle from "sent" through to
+"received". Custom clinic module.
+
+## Public API
+
+Routes mounted at `/api/v1/lab-orders/`.
+
+- `GET    /lab-orders`          — list, filterable by patient/lab/status; `lab_orders.read`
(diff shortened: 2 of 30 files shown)


martinezsalmeron (changes requested):
Thanks for splitting this out, and for fixing the tabs/spaces in `staff_activity/__init__.py` (commit `bfc89f2`) — that one's resolved. RBAC looks solid too: every payroll/staff_activity route has `require_permission`, and bank details stay write-only (`StaffPayrollProfileResponse` only exposes `has_bank_account`/`has_tax_id` booleans, never the underlying fields).

The migration issue from the #153 review didn't get fixed, it got worse: `backend/app/modules/staff_activity/migrations/versions/sact_0001_create_staff_activity_logs.py:18` now sets `down_revision = "meds_0001"` — that revision doesn't exist anywhere in this repo (there's a `media` module, not `medications`, in the tree this branch actually carries). `alembic upgrade heads` will fail outright on this branch. `payroll/migrations/versions/payr_0001_create_payroll_tables.py`'s own docstring says this "had to be manually fixed after the fact" — that fix isn't in the diff. Could you point `sact_0001`'s `down_revision` at `"0001"` (matching `payr_0001`'s own root) so `staff_activity` has its own clean branch?

A few more from the original review that are still open:
- Money math is still `float` throughout `payroll/service.py` (lines 130, 185, 232-234, 254-256) — for payroll specifically I'd push for `Decimal` given rounding errors compound across pay periods.
- `PayrollPeriod.status` exists (`draft`/`processed`/`paid`) but nothing checks it — `generate_entries` can re-run on an already-processed period, and `PayrollEntryService.update` can edit an already-paid entry. Worth a guard on both.
- `StaffPayrollProfileService.create` inserts `data["user_id"]` with no check that the user is actually a member of this clinic.
- `staff_activity`'s handler commits a full raw event payload (`handlers.py:87`, including patient names/phones) synchronously on every platform event, with no retention/purge job anywhere in the module.
- No test files for either module — payroll is financial data, worth at least a happy-path suite before this lands.

Happy to help with the Decimal conversion or the status-guard if useful — the RBAC and encryption work here is solid, it's mostly the migration and the money-math/state-machine gaps standing between this and merge.

---

PR #159, opened 2026-08-05
feat(procurement): Add cleaned supplier data and seeding script

lamanji:
Adds initial supplier data management for procurement and inventory tracking:

Includes cleaned dataset backend/data/suppliers_cleaned.csv.

Adds data seeding script backend/scripts/seed_suppliers.py.


martinezsalmeron (changes requested):
Thanks for splitting this out. The `manifest.depends` declarations across the suite are still accurate (`purchase_orders` → `contacts`/`inventory`/`suppliers`, etc.), and `ruff` only flags one unused import (`PurchaseOrderItem` in `receiving_service.py:28`) — that part's in good shape. But the data-integrity and PII issues from the #153 review carry over unfixed:

- **Inventory stock loses writes under concurrency, still present.** `InventoryService.record_movement` (`backend/app/modules/inventory/service.py:97-158`) still does a plain `select` (line 74-78, no `with_for_update`) then `item.quantity_on_hand = new_quantity` (line 139) — no row lock, no atomic update. This is the method every module funnels stock changes through, including auto-deduction, so concurrent movements silently under-count. Needs `SELECT ... FOR UPDATE` or an atomic `UPDATE ... SET quantity = quantity + :delta`.
- **PO receiving still not atomic**, for the same underlying reason: `receiving_service.py:99-118` loops calling `record_movement`, which commits on every call (`inventory/service.py:157`) and publishes `INVENTORY_MOVEMENT_RECORDED` before the receipt/PO-status update at line 128 — a failure mid-loop leaves stock incremented with no receipt row. PO numbering (`purchase_orders/service.py:53-61`, `count(*)+1`) still collides after a delete+recreate.
- **`backend/data/suppliers_cleaned.csv` is still in the diff and still contains what looks like real third-party PII** — a populated `contact_person` name and phone numbers across multiple rows. This shouldn't be committed to a public repo; please drop the file (and scrub it from the branch history if it's already been pushed).
- **Migration chain is still one thread across five modules** (`inv_0001→...→inv_0003→supp_0001→si_0001→po_0001→po_0002→sr_0001`), all `removable=True`. Each has its own `branch_labels`, but `down_revision` still chains through the previous module's head rather than rooting independently — this is the issue #56 uninstall-safety pattern CLAUDE.md calls out. Uninstalling any one module in the middle orphans everything downstream.
- `inventory_reorder/__init__.py:28` doesn't declare `contacts` in `depends`, but `service.py:80-81` imports `contacts.models.Contact` and does `db.get(Contact, ...)` with no `clinic_id` filter.
- Still zero test files for the procurement chain.

Given the concurrency bug affects live stock counts, I'd treat that one as the blocker to fix first even ahead of the rest. Let me know if it'd help to pair on the row-locking fix.

---

PR #160, opened 2026-08-05
feat(ui): Add navigation configs, permissions, layouts, and notification settings

lamanji:
Enhances core navigation and user interface layout structures:

Adds sidebar navigation (SidebarNav.vue) and navigation groupings (navGroups.ts, settingsSuperGroups.ts).

Updates layout configurations and permission definitions.

Adds notification template management UI (TemplatesSettingsPage.vue) and composables.

Adds migration notif_0004_add_sms_enabled.py for SMS notification flags.


martinezsalmeron (changes requested):
Thanks for splitting this out — the i18n coverage here is clean (both `SidebarNav.vue` and `TemplatesSettingsPage.vue` use `t()` throughout, no hardcoded English strings), and the `notifications` migration (`notif_0004_add_sms_enabled.py`) chains correctly off `notif_0003` with no new branch label, so no issue there.

One regression from the #153 review is still present and would break existing users: `frontend/app/config/permissions.ts` repoints the `documents` key from `media.documents.read`/`write` to `documents.read`/`write`. That key is already consumed by `media/components/{DocumentCard,DocumentGallery,PhotoGallery}.vue` and `patients/components/patient/AdministrationTab.vue` on `main`, and their backend routes still enforce `media.documents.*`. Once this merges, a non-admin with `media.documents.*` granted would see those existing buttons hidden or 403ing. Since the new `documents` module (in #155) needs its own permission key anyway, could it get a distinct one (e.g. `documents.generate`) instead of reusing/repointing the existing `documents` entry?

Separate, smaller thing I noticed while checking: the new `TemplatesSettingsPage.vue` / `useNotificationTemplates.ts` don't gate their create/edit/delete actions behind `can(PERMISSIONS.x.y)` at all — worth adding given every other settings page in the app does.

Also flagging for awareness, not blocking: `frontend/Dockerfile.prod` bumps Node 20→22-alpine and `@nuxtjs/i18n` goes `^9.5.0`→`^10.6.0` (major version). Worth a manual smoke test of the i18n locale switching after that bump, since majors sometimes change runtime config shape.

---

PR #163, opened 2026-08-08
fix - alembic-sigill-arm64-docker-cryptography-47

tresundios:
fix: pin cryptography<47 and add @backend-modules alias for agenda tests

- Pin cryptography to <47 to avoid SIGILL on ARM64 Docker.
- Use @backend-modules alias in agenda tests instead of relative paths.
- Configure @backend-modules in vitest.config.ts for Docker vs local.


martinezsalmeron:
Thanks for the PR! I couldn't reproduce the SIGILL, so I'd like more details before pinning.

**Repro attempt:** on Apple Silicon (aarch64, native arm64 Docker image), `cryptography 50.0.0` imports and runs AES-GCM without issues — three majors past the proposed `<47` ceiling. So this doesn't look like a general ARM64 problem.

Could you share:
- Host arch/OS and Docker setup (is the backend image being emulated, e.g. an amd64 image under QEMU/Rosetta?)
- Base image and exact cryptography version that crashes
- The SIGILL backtrace
- Whether it also reproduces with 48.x / 49.x / 50.x

**On the pin itself:** capping a security-critical library below 47 indefinitely means missing upstream security patches. If the crash turns out to be emulation- or CPU-specific, the right fix is on the image/platform side (or a documented workaround), not a global upper bound.

**On the vitest alias change:** this is unrelated to the SIGILL fix, and the relative import it replaces already works locally, in CI and in Docker (that's what the comment it deletes explains). `@backend-modules` also isn't declared in tsconfig/ESLint, so typecheck/lint would flag it. Please drop it from this PR.

Happy to revisit with a repro. Without one I'll close this in a couple of weeks.

tresundios:
I have removed frontend updates.  

Here is the requested information below,  let me check and update here for 48.x / 49.x / 50.x after verification.

=== Host OS ===
ProductName:		macOS
ProductVersion:		26.2
BuildVersion:		25C56
=== Architecture ===
arm64
Darwin Mac-mini.local 25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:08:48 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T8132 arm64
=== Mac hardware ===
\Hardware:

    Hardware Overview:

      Model Name: Mac mini
      Model Identifier: Mac16,10
      Model Number: MCYT4HN/A
      Chip: Apple M4
      Total Number of Cores: 10 (4 performance and 6 efficiency)
      Memory: 24 GB
      System Firmware Version: 13822.61.10
      OS Loader Version: 13822.61.10
      Serial Number (system): K6P57CX2J9
      Hardware UUID: 94448AFE-625B-50BA-B95A-7458E2161B93
      Provisioning UDID: 00008132-000919D201F9001C
      Activation Lock Status: Disabled

=== Docker ===
Client:
 Version:           27.3.1
 API version:       1.47
 Go version:        go1.22.7
 Git commit:        ce12230
 Built:             Fri Sep 20 11:38:18 2024
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.35.1 (173168)
 Engine:
  Version:          27.3.1
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.7
  Git commit:       41ca978
  Built:            Fri Sep 20 11:41:19 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.21
  GitCommit:        472731909fa34bd7bc9c087e4c27943f9835f111
 runc:
  Version:          1.1.13
  GitCommit:        v1.1.13-0-g58aa920
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
=== Docker client/server architecture ===
Client: darwin/arm64 | Server: linux/arm64
=== Docker images/containers ===

REPOSITORY                    TAG         CREATED
dentalpin-backend             latest      23 hours ago
dentalpin-frontend            latest      24 hours ago
python                        3.12-slim   3 weeks ago
postgres                      16-alpine   4 weeks ago
redis                         7-alpine    3 months ago
postgres                      15-alpine   3 months ago

NAMES                  IMAGE                STATUS
dentalpin-frontend-1   dentalpin-frontend   Up 3 minutes (healthy)
dentalpin-backend-1    dentalpin-backend    Up 3 minutes (healthy)
dentalpin-db-1         postgres:15-alpine   Up 3 minutes (healthy)

=== Docker image architecture ===
Image=dentalpin-backend:latest Architecture=arm64 OS=linux Variant=v8
Image=dentalpin-frontend:latest Architecture=arm64 OS=linux Variant=v8
Image=fastapi-app:latest Architecture=arm64 OS=linux Variant=
Image=python:3.12-slim Architecture=arm64 OS=linux Variant=v8
Image=postgres:16-alpine Architecture=arm64 OS=linux Variant=v8
Image=redis:7-alpine Architecture=arm64 OS=linux Variant=v8
Image=postgres:15-alpine Architecture=arm64 OS=linux Variant=v8

tresundios:
I completed the requested verification inside the `dentalpin-backend` container on `aarch64` using a `cryptography` import test and AES-GCM test.

Results:

- `46.0.7`: OK
- `47.0.0`: SIGILL / exit code 132
- `48.0.0`: SIGILL / exit code 132
- `49.0.0`: SIGILL / exit code 132
- `50.0.0`: SIGILL / exit code 132

Thus, in this specific container environment, all tested versions from `47.x` through `50.x` crash with an illegal instruction. This differs from the Apple Silicon environment where `50.0.0` succeeds, so the issue may depend on the wheel build, base image, OpenSSL build, Docker runtime, or exposed CPU features.

For now, `cryptography<47` is required for this deployment environment. I suggest documenting this as an environment-specific workaround while investigating the underlying compatibility issue, rather than treating it as a general ARM64 limitation. [web:3]

martinezsalmeron (commented):
Thanks for the thorough follow-up — this is exactly the kind of repro I was asking for.

Confirmed the vitest/`@backend-modules` change is gone; the diff is now the single line in `backend/pyproject.toml`. Appreciate you scoping it back down.

On the SIGILL data itself: this actually sharpens my original concern rather than resolving it. Your matrix shows `46.0.7` fine and `47.x`–`50.x` crashing **inside the `dentalpin-backend` container**, but `50.0.0` running clean on the **bare macOS host** (same chip, same host CPU). Same silicon, same cryptography build era, different result depending on whether it's running under Docker Desktop's Linux VM. That points at the virtualization/hypervisor layer exposing CPU features (likely AArch64 crypto extensions surfaced to the guest via CPUID) that the newer Rust-backed `cryptography` code path detects and uses, but that Docker Desktop's VM doesn't actually execute correctly — not a general "cryptography ≥47 is broken on ARM64" problem.

If that's right, `cryptography<47` in `pyproject.toml` fixes it for every Docker Desktop on Apple Silicon dev environment, but also permanently blocks security patches for everyone else who doesn't hit this (native linux/arm64 hosts, CI runners, production deployments) — indefinitely, since there's no tracking issue to unpin against.

Before merging a floor like this I'd want to see:
1. An upstream report to [pyca/cryptography](https://github.com/pyca/cryptography/issues) with exactly this comparison (host vs. Docker Desktop VM, the 46→50 matrix) — `cryptography`'s Rust backend has had SIGILL-under-virtualization reports before, and they're best positioned to say whether this is a known CPU-feature-detection bug and whether there's a runtime workaround (e.g. an env var to disable the accelerated path) instead of a version ceiling.
2. If no workaround exists, consider scoping the pin narrower than a global lower bound — e.g. documenting it as a known Docker Desktop / Apple Silicon caveat with a workaround, rather than capping the dependency for all platforms.

Not blocking on my own authority — flagging for @martinezsalmeron to weigh in on whether the global pin is acceptable as a stopgap or whether this needs the upstream report first.

---

martinezsalmeron:
Thanks for the version matrix — that gave me exactly what I needed to try a side-by-side repro. Result: **I can't reproduce the SIGILL on any version**, in the same `dentalpin-backend` image, also on Apple Silicon Docker Desktop.

**My test:** throwaway containers from the same `dentalpin-backend` image (linux/arm64), `pip install cryptography==<version>`, then import + AES-GCM encrypt/decrypt round-trip:

| Version | Result |
|---------|--------|
| 47.0.0 | OK |
| 48.0.0 | OK |
| 49.0.0 | OK (also running in my long-lived backend container, 29h uptime, healthy) |
| 50.0.0 | OK |

**The relevant difference between our environments isn't the chip — it's the Docker Desktop version:**

| | Yours | Mine |
|---|---|---|
| Chip | Apple M4 | Apple M5 |
| Docker Desktop | 4.35.1 (engine 27.3.1, **Sep 2024**) | current (engine 29.5.3) |
| Result on 47–50 | SIGILL | OK |

Docker Desktop 4.35.1 predates the M4 launch. There's a well-documented family of SIGILL crashes in containers on M4/M5 hosts caused by the VM layer mishandling/exposing the ARMv9 SME2/SVE CPU features to the Linux guest — it hit the JVM ([corretto-21#85](https://github.com/corretto/corretto-21/issues/85)), .NET ([dotnet/runtime#122608](https://github.com/dotnet/runtime/issues/122608)), podman ([containers/podman#28312](https://github.com/containers/podman/issues/28312)), and any native code doing runtime CPU-feature detection — which is exactly what `cryptography`'s Rust backend does more aggressively from 47 onward. Newer Docker Desktop releases fixed this. My M5 has the same SME2 extensions as your M4, and with a current Docker Desktop nothing crashes — which fits the hypervisor-bug explanation and rules out "cryptography ≥47 is broken on ARM64 Docker".

**Could you upgrade Docker Desktop to the latest release and re-run your 47–50 matrix?** I expect the SIGILL to disappear.

Given that the crash doesn't reproduce on an up-to-date environment, I'm not going to merge a global `<47` cap on a security-critical library — it would silently block security patches for every environment that isn't affected (CI, native arm64 hosts, production). If the crash *does* survive a Docker Desktop upgrade, the right next step is an upstream report to [pyca/cryptography](https://github.com/pyca/cryptography/issues) with the host-vs-VM comparison, and we can revisit.

I'll close this PR once you confirm the upgrade fixes it (or if there's no follow-up in a couple of weeks).

---

tresundios:
After upgrading Docker Desktop Engine 29.6.2, the SIGILL disappears across all four cryptography versions.

Matrix results (Docker engine 29.6.2, Apple Silicon M4)
----------------------------------------------------

Version | Result
-- | --
47.0.0 | OK
48.0.0 | OK
49.0.0 | OK
50.0.0 | OK


Test used the same `dentalpin-backend:latest` image (`linux/arm64`) rebuilt from the current `main` branch. Each version was installed in a throwaway container and exercised with `AESGCM` import \+ encrypt/decrypt round-trip.

I think you can close the PR.

Apologize, that you spend on this, let me work on some other issue

martinezsalmeron:
Great, thanks for running the full matrix down and confirming — glad it turned out to be the Docker Desktop/M4 hypervisor bug rather than something in `cryptography` itself. Closing this out. No apology needed — this is exactly the kind of report that's useful to have on record for the next person who hits it on an old Docker Desktop build.

Appreciate you sticking with the back-and-forth. Feel free to jump on any of the `good first issue` / `help wanted` labelled issues if you're looking for the next thing.

---

PR #189, opened 2026-08-18
feat(recall_reminders): notify patients when a recall is created

lamanji:
Starting small per your suggestion — recall_reminders, since it turned out to
have the smallest surface (no new tables/migrations, both its dependencies
are already on main).

What's here:
- The module itself (was already written, just never actually wired up —
  it had no entry point registered in pyproject.toml or frontend/modules.json,
  so the plugin loader never discovered it at all)
- A real bug I found while testing: NotificationGateway.enqueue() only
  flushes by design (its own docstring says the caller owns the commit),
  but the handler never called db.commit(). Every enqueued reminder was
  silently rolled back when the session closed — no recall reminder could
  ever have actually been sent. Fixed.
- Two tests: tenant-isolation happy path, and idempotency on the dedup_key
  (same recall firing twice shouldn't double-send).

Verified clean on a fresh checkout of main: migrations apply, app imports,
both tests pass.


martinezsalmeron (commented):
Re-reviewed after the latest push (`9d90352`). Ran `ruff check` / `ruff format --check` locally over the touched files — both clean, confirming that part of your report. (This environment has no docker/Postgres available to me, so I couldn't re-run the full pytest/`check_docs_coverage.py` suite myself, but the diff matches what those would expect.)

Everything from the previous round looks properly addressed:
- Handler is transactional per ADR 0019 (`db.begin_nested()`, `recalls/service.py:298` now passes `db=db`), and the `db.commit()` bug is gone.
- Template shipped for all 5 locales instead of requiring per-clinic setup.
- `rr_0001` gives it an isolated Alembic branch, and the `alembic.ini` `version_locations` fix is a good catch — that static list is easy to miss.
- Docs, changelog, and catalogs all regenerated correctly. `docs/events-catalog.md`'s `recall.created` row now lists `recall_reminders` as a subscriber.

One very minor, non-blocking nit: `docs/technical/recall_reminders/permissions.md` wasn't added. `check_docs_coverage.py` only requires it when a module returns permissions from `get_permissions()` (this one returns `[]`), so CI won't fail on it — but `CLAUDE.md`'s new-module checklist lists `{overview,events,permissions}.md` as a set. Worth a one-line file noting "no permissions" for consistency with other modules, but I wouldn't hold up the merge for it.

Flagging your own callout again so it doesn't get lost: the Tamil (`ta`) template wording should get a native-speaker check before this ships.

Thanks for the very thorough turnaround — this is a good, clean addition. @martinezsalmeron, this reads ready to merge once the Tamil copy is confirmed (or dropped to a fallback locale) if you want to be strict about that.

---

PR #208, opened 2026-08-19
feat(patient_admin): patient-to-patient relationships (Lien de Parentee)

lamanji:
Adds the patient_admin module: directed patient-to-patient links (parent/child, spouse, sibling, guardian/ward, other) surfaced inline on the patient page. Registered as a `dentalpin.modules` entry point in `pyproject.toml` and added to `frontend/modules.json` (one line each, same as the recall_reminders PR).

## Security & Correctness Fixes

Two fixes applied on top of the existing implementation:

1. **`get_relationship()` had no `clinic_id` filter** , same bug class as the documents module fix from earlier. Not currently exploitable here (update/delete both additionally require the URL's `patient_id`, already verified to belong to the caller's clinic, to match the relationship row , see `CLAUDE.md` for the full reasoning on why that makes it safe today but fragile for tomorrow), but a future caller that reaches `get_relationship` directly shouldn't have to know that convention to stay safe. Added the filter; both router call sites now pass `clinic_id`.

2. **`padm_0001`'s migration docstring claimed "no dependency"** , wrong, because it FKs to `patients.id`, and patients lives on its own unlabeled chain separate from `patient_admin`'s branch. Added `depends_on = ("pat_0003",)` ,same pattern used in `documents/doc_0001` , and corrected the docstring. Verified `alembic upgrade heads` succeeds across 3 different `PYTHONHASHSEED` values on a fresh database.

## CI-Gate Lessons (applied proactively)

Applied the same CI-gate lessons the `recall_reminders` PR's review surfaced:

- Added `patient_admin`'s `migrations/versions` dir to `alembic.ini`'s static `version_locations` list , the manifest validator's branch-isolation check loads its ScriptDirectory from `alembic.ini` directly (a different path from the dynamic discovery the live `alembic upgrade heads` CLI uses). Confirmed `tests/test_manifest_validator.py` fails without this line and passes with it.
- Added `docs/technical/patient_admin/overview.md` and `permissions.md` (`get_permissions()` returns 2 permissions, so `permissions.md` is required , confirmed via `scripts/check_docs_coverage.py --strict`, which now exits 0).
- Dropped the "Clinic Custom" author placeholder (repo convention is "DentalPin Core Team") , same fork-specific leftover caught in `recall_reminders`.
- `ruff format/check` clean on every touched file.

## Testing & Documentation

- Added `CLAUDE.md` and `CHANGELOG.md` (module had neither , note these are separate from `docs/technical/patient_admin/`, which is what `check_docs_coverage.py` actually validates).
- Added `tests/modules/patient_admin/test_tenant_isolation.py`: a happy-path create+list test, and a tenant-isolation test for `get_relationship` confirmed to fail against the unfixed code and pass against the fix.

## Verification

Verified end-to-end on a clean `upstream/main` checkout:
- `alembic upgrade heads` succeeds across 3 hash seeds
- The app imports cleanly
- The module is discovered by the plugin loader
- `ruff` is clean
- `check_docs_coverage.py --strict` exits 0
- All 9 relevant tests pass (2 in this module, 7 in the manifest validator)


martinezsalmeron (commented):
Thanks for this, @lamanji — welcome, and nice first PR. Checked it against `CLAUDE.md`'s module rules and ran the gates I could locally (no docker/Postgres in this environment, so no pytest, but `ruff check` / `ruff format --check` on `backend/app/modules/patient_admin` and its tests are both clean).

**What's solid:**
- `depends: ["patients"]` declared and matches the actual cross-module read (`Patient.full_name`) — legal under ADR 0002.
- Own Alembic branch (`branch_labels = ("patient_admin",)`, `depends_on = ("pat_0003",)` for the FK to `patients.id`), and `backend/alembic.ini`'s `version_locations` updated — that's a step people miss.
- Every query is `clinic_id`-scoped, including the direct-`relationship_id` lookup in `service.py`'s `get_relationship` — and `test_tenant_isolation.py::test_get_relationship_is_clinic_scoped` actually proves it cross-clinic, not just cross-patient.
- Docs are complete: `CLAUDE.md`, `CHANGELOG.md`, `docs/technical/patient_admin/{overview,permissions}.md`, catalog regenerated.
- Self-caught the `padm_0002` exemption-status removal and explained why (APCI is a computed flag, not manual entry) rather than leaving dead scope in.

**One gap worth fixing before merge:** the RBAC is real on the backend (`require_permission("patient_admin.relationships.write")` on the mutating routes) but isn't wired on the frontend. `frontend/app/config/permissions.ts` has no `patient_admin` entry, and `PatientAdminCard.vue`'s edit pencil, the add-relationship form, and the remove buttons all render unconditionally — a `hygienist` (whose manifest grant is `read`-only) would see the full edit UI and get a 403 only after clicking. Per `CLAUDE.md`'s RBAC section, add `patientAdmin.relationships.{read,write}` to `permissions.ts` and gate those controls with `usePermissions().can(...)`, the way `patients_clinical` does it for its own write actions.

Two small non-blocking nits:
- `service.py`'s `update_relationship` does `setattr(row, key, value)` only `if value is not None` over a `model_dump(exclude_unset=True)` payload — that means a client can never explicitly clear `notes` via `PUT`, only leave it unset. Minor, but worth a conscious call either way.
- `padm_0001` creates `patient_admin_exemption_status` and `padm_0002` drops it two commits later — since neither migration has shipped in a release yet, squashing into a single `padm_0001` without the table would leave a cleaner history. Not worth blocking on, your call.

Nothing here is a redesign — the permissions-wiring gap is the only one I'd want closed before this merges. Thanks again for a well-tested, well-documented module.

---

PR #210, opened 2026-08-19
feat: add India GST invoicing compliance module

tresundios:
## Summary
- Adds the `india_gst` module to provide India-specific GST invoicing compliance for Indian dental clinics.
- Supports GSTIN capture, place-of-supply rules, SAC defaults, CGST/SGST/IGST tax splits, GST document numbering, credit-note reversal, and reconciliation reporting.
- Extends billing through the country-gated compliance-hook architecture, without changing the core billing schema.

## Modules touched
- New: `india_gst`
- Updated: `billing`, PDF invoice generation, frontend invoice views, localisation, migrations, module catalogs, Docker font dependencies

## Key capabilities
- Clinic GST profile and supplier GSTIN configuration.
- Recipient GSTIN stored through the existing billing tax-ID field.
- Automatic intra-state versus inter-state determination based on place of supply, with CGST+SGST or IGST breakdowns.
- SAC-code defaults for treatment catalog items, including a safe one-click auto-configuration for missing defaults.
- Financial-year-scoped GST document numbering (April–March).
- GST reconciliation summary, transaction report, and CSV export.
- GST-specific invoice UI panels, status badges, filters, and Indian-clinic labels.
- Invoice PDF enhancements with GST breakdown, supplier/recipient GSTINs, place of supply, GST label overrides, and Tamil PDF support.
- Credit notes preserve the original invoice’s place of supply and correctly reverse GST amounts.

## E-invoice scope
- Includes e-invoice data structures, settings, UI states, and status endpoints.
- Live GSP/IRP provider submission is intentionally not included in v1; retry returns `409` rather than reporting a fabricated submission success.

## Limitations
- v1 calculates GST only for clinics with `registration_type = "regular"`.
- Composition, unregistered, and exempt registration types can be stored but do not yet apply alternate tax rules.

## Safety
- The module is visible and active only for clinics configured with country `IN`.
- Uninstallation is blocked when issued/non-draft invoices contain GST line-item data, preventing loss of historical tax breakdowns.
- GST compliance snapshots are stored at invoice issue time so later clinic-setting changes do not alter historical invoices.

## Test plan
- [ ] Configure an Indian clinic with a regular GST registration and supplier GSTIN.
- [ ] Add SAC defaults to treatments, create both same-state and interstate draft invoices, and verify CGST/SGST versus IGST splits.
- [ ] Issue an invoice and verify GST document number, compliance snapshot, UI badge, and PDF GST section.
- [ ] Create a credit note and verify it inherits the original place of supply and reverses GST correctly.
- [ ] Verify Tamil invoice PDF rendering.
- [ ] Verify GST reconciliation reports and CSV export.
- [ ] Confirm e-invoice retry returns the expected `409` while no provider is configured.
- [ ] Attempt to uninstall after issuing a GST invoice and confirm the uninstall guard prevents it.


martinezsalmeron (commented):
Hi @tresundios — thank you for this contribution. The module core is genuinely well built: it clones the verifactu hook architecture faithfully (billing has zero backend knowledge of india_gst), the snapshot-at-issue and credit-note reversal logic are correct and well tested, the migration sits cleanly on its own Alembic branch, and the slot-based frontend integration is exactly how we want modules to extend billing. The bilingual documentation and the honest 409 on e-invoice retry are appreciated too.

That said, the review found issues that block a merge as-is, and rather than send it through several review round-trips we've finished it ourselves: **9 fix-up commits pushed directly to this PR's branch** (via maintainer edits) — all your commits and authorship are preserved, our fixes land on top. Summary of what changed and why:

**The red e2e was self-inflicted, not the module.** The failures were in specs the module doesn't touch. Root cause: the branch rewrote the CI e2e job from `nuxt dev` to a production build (which exposes an i18n lazy-load race with text-based locators) and added a global `vue-router: 5.0.4` override — a major version `@nuxtjs/i18n` doesn't support — with the lockfile left out of sync (`npm ci` fails). We reverted all infra unrelated to the module (CI job, router pin, devtools defaults, `modules.json` regeneration, the demo-persona change, and the three loosened e2e specs); the prod-build e2e idea is rescued as its own issue. Please keep future module PRs free of host-infra changes — they're impossible to review together.

**Blockers fixed on top of your work:**
1. *Lifecycle*: `install()` ran the demo seed against every real clinic with `country=IN` — overwriting `trade_name`, stamping a placeholder GSTIN, re-taxing already-issued invoices and creating fake invoices against real patients; `uninstall()` swept a destructive cleanup across all clinics. Issued invoices are immutable fiscal documents. Install now only registers the hook; the seed is gone (the auto-configure action covers real setup).
2. *Duplicate GST numbers*: the document number reused billing's `sequential_number`, which resets on the calendar year — inside one Indian FY (Apr–Mar), January restarts at 0001 and `GST/FY26-27/0001` gets issued twice (Rule 46(b) violation, nothing enforced uniqueness). There's now a per-(clinic, prefix, FY) counter table with a unique constraint and `SELECT … FOR UPDATE`.
3. *Multi-tenancy*: 7 queries looked rows up by foreign id without `clinic_id` (hook, draft-update, reports/CSV) — cross-tenant reads. All filtered now, with a multi-tenant test suite that fails against the old code.
4. *PDF HTML injection*: the hook interpolated user-controlled values (GSTINs, trade name) into raw HTML that billing spliced into WeasyPrint unescaped. The contract is now structured data (`compliance_section` rows) that billing renders and escapes.
5. *CGST ≠ SGST*: the remainder-absorption split produced 50.00/50.01 — both heads are levied at the same rate on the same value and must be equal (and rounding is now HALF_UP, not banker's).
6. Also: `clinic_state` is now required before issuing (missing state silently taxed everything as inter-state IGST), `settings.read` granted to clinical roles (the invoice panel 403'd for receptionists), the CSV export uses an authenticated fetch (`window.open` can't send the JWT — it always got 401), formula-injection guard on the CSV, real SQL pagination, and the isolation-test whitelist entries are gone (the undeclared budget/patients imports lived in the deleted seed).
7. We removed the unreachable e-invoice scaffolding (provider ABC with no implementation, a queue nothing drains, columns nothing writes, an inert rounding setting). The applicability tracking (`not_required`/`not_configured`) and the honest 409 stay. When a real GSP/IRP adapter lands, its columns and provider config land with it.

**Three domain questions where your India GST knowledge would really help:**
1. Rounding: we settled on CGST = SGST = half the line tax, each rounded HALF_UP per head (so an odd-paise line can drift ±1 paisa vs the line total). Does that match what GSTR-1 filing tools expect in practice, or do clinics' accountants prefer invoice-level rounding under Section 170?
2. FY numbering: serials now restart at 1 every April 1 per series prefix. Any reason a Tamil Nadu practice would want a continuous series across FYs instead?
3. E-invoice threshold: v1 compared a single invoice's total against `turnover_threshold`, but the statutory trigger is aggregate annual turnover. We've left the field as-is and opened a follow-up — how do the clinics you know actually determine applicability?

Thanks again — the module will ship, and the GST core you wrote survives essentially intact.

martinezsalmeron (approved):
Fix-up validated end to end: multi-tenant + permission suites, FY sequencing, PDF escaping, acceptance matrix and the e2e root-cause fix are all green. Demo fixture and turnover-based e-invoice applicability landed cleanly in the follow-ups. Shipping DentalPin's first India-compliance module — thank you @tresundios! 🇮🇳

Issue #65, opened 2026-04-28
Public integrations API + Zapier app (also powering Make / n8n / custom)


martinezsalmeron:
## Problem

Every clinic runs DentalPin alongside a stack of other tools — accounting (Holded, Quipu, Contasimple, QuickBooks), email marketing (Mailchimp, Brevo, Customer.io), payments (Stripe, Bizum, PayPal), website lead forms, ad platforms (Meta Ads, Google Ads), reviews (Google, Trustpilot, Doctoralia), team chat (Slack, Discord, Telegram), BI (Sheets, Looker Studio, Metabase), CRMs (HubSpot, Pipedrive). They all need to exchange data with DentalPin.

Building a bespoke integration for each is a treadmill we cannot win. **Zapier (and its peers Make and n8n) is the generic glue** — if DentalPin speaks their language, every clinic can wire whatever they want, in minutes, with no code from us.

A clinic owner asked for a Zapier integration. The right answer is bigger: build the **public API + webhook subscription surface** that powers Zapier *and* Make *and* n8n *and* any custom integration, then publish a thin Zapier app on top.

## Why this is high-leverage

What clinics will build, day one, with no code from us:

**Outbound (DentalPin → world)**
- New patient → add to Mailchimp / Brevo / Customer.io list.
- Appointment completed → create invoice line in Holded / Quipu / Contasimple / QuickBooks.
- Treatment plan accepted → Slack/Discord ping the team; create Stripe payment link; trigger DocuSign for consent.
- Treatment plan rejected → enqueue nurture sequence in HubSpot / Pipedrive.
- Payment received / failed → Sheets log; Slack alert; receipt in accounting tool.
- Recall due → marketing email; WhatsApp via the notifications gateway (#63).
- Patient birthday → send card via Mailchimp.
- Daily metrics → Google Sheets / Looker dashboard.
- Review request 24 h after a finished appointment → Trustpilot / Google Reviews flow.
- High-value patient activity → CRM enrichment.

**Inbound (world → DentalPin)**
- Website contact form → create lead → create patient on intake.
- Meta / Google Ads lead → create patient + recall to call this month.
- Calendly / external booking → create appointment.
- Stripe / Bizum / PayPal payment → log payment + close the related plan item.
- Google / Trustpilot review posted → log on patient timeline.
- WhatsApp reply via BSP → log on patient timeline (already covered by #63 inbound, but Zapier route works for clinics on free tier).
[...]

PR #246, opened 2026-08-22
feat(integrations): webhook subscriptions with outbox delivery (issue #65 Phase 1)

hirad121:
@C:\Users\hirad\AppData\Local\Temp\claude\C--Users-hirad-OneDrive-Desktop-Projectssssss-contribute\8c6932d5-9039-42f5-b8d2-fe5841d7cdd6\scratchpad\pr246-new-body.txt


martinezsalmeron (changes requested):
Thanks — this is the right slice of #65, and the code is solid: the outbox mirrors `notifications.gateway` faithfully, module isolation is clean, migration on its own branch, docs/catalogs/CHANGELOG all present. I ran the 45 tests on a fresh DB (all green), `ruff` clean, `generate_catalogs --check` clean, and poked the SSRF guard with `[::ffff:169.254.169.254]`, `[::ffff:127.0.0.1]`, `0.0.0.0`, `10.0.0.1` — all blocked. The uninstall round-trip fails locally for me exactly like `patient_relationships`' does (env issue), so CI decides that one.

Requesting changes for two blockers and five small items.

## Blockers

1. **`socket.getaddrinfo` is synchronous inside the event loop** (`url_safety.py::_resolved_ips`). It runs from a Pydantic `field_validator` on `POST/PATCH /webhooks/subscriptions` (request path) and from `client.post_webhook` (scheduler tick, same loop as the API). A non-resolving host or a slow resolver freezes the whole API for seconds per call. Fix: `await asyncio.get_running_loop().getaddrinfo(...)`, and move the URL check out of the Pydantic validator into the service (validators can't await). Please add a test covering it.

2. **References to `notes/dentalpin/65-integrations-api.md` and "approved by Ramón 2026-08-21"** in 10+ places (`__init__.py`, `models.py` — including a wiki-link to `../../../../../notes/`, `gateway.py`, `handlers.py`, `router.py`, `service.py`, `signing.py`, `triggers.py`, `url_safety.py`, `CLAUDE.md`, `CHANGELOG.md`). Those are dead links to your private notes plus conversation transcript in code. Replace with `issue #65` / ADR 0019 references and drop the names and dates.

## Small, same PR

3. `url_safety.py` claims to defend against DNS rebinding. It doesn't: we validate, then httpx resolves again (TOCTOU). Fine for Phase 1, but the docstring should say "narrows the window", not "defends". Pinning the resolved IP in the transport is a follow-up.
4. `update_subscription` uses `data.get(field) is not None`, so `description` can never be cleared to `null`. The router already passes `exclude_unset=True`; use `if field in data`.
5. Webhook body has no `occurred_at` timestamp and no event id stable across subscribers (`delivery_id` is per subscription). Add at least `occurred_at`; frozen sample payloads per trigger (#65 §3) can stay a follow-up but please note it in the module CLAUDE.md.
6. Add an explicit `get_tools()` returning `[]` (CLAUDE.md: mandatory even if empty).
7. Rename the header to `X-DentalPin-Signature`. It's a public contract; renaming later is breaking.

## Non-blocking, want your opinion

- Auto-disable after 10 consecutive failures with 5 attempts per delivery: an endpoint down for ~1h gets disabled and nobody finds out (no UI, no event). Zapier is more tolerant. Either publish a `webhook.subscription.disabled` event on the bus or raise the threshold until the admin UI exists.
- `depends: ["patients"]` isn't needed just to consume its event — harmless, your call.

## Scope

Yes to your offer: keep `appointment.completed` (and API tokens) out of this PR, separate PRs. Let's be explicit in the description that Phase 1 as shipped is infrastructure — with a `{"patient_id","clinic_id"}` payload and no public read API, a Zap can't do anything useful yet.

martinezsalmeron (approved):
All review items addressed — 59 tests green on a fresh DB here, ruff/catalogs/docs checks clean, and I ran the uninstall round-trip myself on a fully migrated DB (CI doesn't run on the fork branch): passes. Merging.

PR #247, opened 2026-08-22
fix: reject weak SECRET_KEY at startup in production (GHSA-hcg9-cm67-2g8f)

hirad121:
## Summary

- Fixes GHSA-hcg9-cm67-2g8f (advisory triage, unmerged): `SECRET_KEY` had no length/strength validator — Pydantic only required non-empty, so a weak key (e.g. 1 char) booted with no error and no warning.
- `SECRET_KEY` signs JWTs and derives the Fernet key used to encrypt SMTP passwords and Veri*Factu tax certificates at rest, so a weak value compromises all three at once.
- Adds a `model_validator(mode="after")` on `Settings`: requires ≥32 chars (matches `.env.example`'s `openssl rand -hex 32`), fails hard only when `ENVIRONMENT=production`, warns otherwise — per your go-ahead in the "Interested in contributing" email thread. Separate, decoupled from #246/#65 as requested.

## Modules touched

- none (core `app/config.py`)

### Always

- [x] `cd backend && ruff check . && ruff format --check .` passes
- ~~`cd frontend && npm run lint` passes~~ (no frontend changes)
- [x] Tests added or updated when behaviour changed

### If a new module was added

~~N/A — no new module~~

~~N/A — no events touched~~

~~N/A — no permissions touched~~

~~N/A~~

~~N/A — small, self-contained validator, not an architectural change~~

~~N/A~~

~~N/A — `app/config.py` isn't inside a module~~

## Test plan

- [x] `backend/tests/test_secret_key_strength.py` (new, 5 tests): weak key + `ENVIRONMENT=production` raises; weak key + `development`/`test` warns and boots; strong key + `production`/`development` boots with no warning. Confirmed to fail on pre-fix `config.py` (import error — validator doesn't exist) and pass on the fix.
- [x] Direct `Settings()` boot checks against every real value CI actually uses (`test-secret-key-for-ci`+`test`, `ci-stub`+`test`, `ci-secret-key-for-e2e`+`development`) — all warn only, none break, since none set `ENVIRONMENT=production`.
- [x] `ruff check` / `ruff format --check` clean on full `backend/`.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -1,6 +1,11 @@
 """Application configuration via environment variables."""
 
+import warnings
+
+from pydantic import model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
+
+MIN_SECRET_KEY_LENGTH = 32
 
 
 class Settings(BaseSettings):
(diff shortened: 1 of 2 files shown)


martinezsalmeron (approved):
LGTM. Verified prod SECRET_KEY length (96) so the production gate won't trip on deploy. Thanks!

PR #364, opened 2026-09-02
fix: realistic occlusal view + crown profile without root

### LINKED ISSUE #48: Realistic occlusal view + crown profile without root
Hi! First of all, great job on the application. 🦷

When you have time, we'd love to hear your thoughts on two small
visual improvements:

**Occlusal view** — the tooth as seen from above is currently shown
as a circle divided into 5 sectors. It would be great to have a shape
closer to the real one (which varies quite a bit between incisors,
premolars and molars) and possibly bring the zones from 5 to 17,
which is the ISO/FDI standard used clinically.

**Crown profile** — it would also be useful to have a lateral view
of the tooth without the root, to use in contexts where space is
limited or where the roots don't add relevant information
(such as in the surface selection popup).

No rush at all, it's more of a future idea than an immediate priority.

Thanks a lot

Mr-Neutr0n:
Fixes #48

Replaced the generic circular occlusal view with anatomically distinct incisor, canine, premolar, and molar outlines (still 5 surfaces); crown-only lateral view in the surface popup was already present; 17-zone expansion was not included as it requires a larger backend change.

No local test suite detected.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/backend/app/modules/odontogram/CHANGELOG.md
+++ b/backend/app/modules/odontogram/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 - feat(i18n): the frontend layer's directional spacing, borders, text alignment and inset positioning now resolve against the document direction (physical→logical CSS utilities, Arabic RTL support).
 
+- feat(#48): realistic per-category occlusal outlines (incisor / canine / premolar / molar) with cusp-aware silhouettes; surface fills clipped to the outline; distal (D) sector is a full quad; molar fissures shortened to the central O so they no longer read as a missing-tooth overlay.
 
 - fix(#101): the module's frontend adopts the useApi error contract — 400/409/422 failures the UI used to swallow now toast the backend's message; calls whose surrounding code already presents the error pass `errorToast: false` (single toast), and hand-built error reads use the shared `errorMessage`/`errorDetail` helpers.
 
--- a/backend/app/modules/odontogram/frontend/components/odontogram/SurfaceSelectorPopup.vue
+++ b/backend/app/modules/odontogram/frontend/components/odontogram/SurfaceSelectorPopup.vue
@@ -68,16 +68,25 @@ const occlusalGroupTransform = computed(() => {
 
 // Get correct X position for M label based on quadrant
 // M (mesial) is always toward midline: left side of view for right teeth, right side for left teeth
-const mLabelX = computed(() => isLeftSide.value ? 43 : 4)
-// Get correct X position for D label based on quadrant
-const dLabelX = computed(() => isLeftSide.value ? 4 : 43)
-
-// Get correct Y position for V label based on arch
-// V (vestibular) faces outward: top for upper teeth, bottom for lower teeth
-const vLabelY = computed(() => isLower.value ? 47 : 8)
-// Get correct Y position for L label based on arch
-// L (lingual) faces inward: bottom for upper teeth, top for lower teeth
-const lLabelY = computed(() => isLower.value ? 8 : 47)
+// M/D/V/L/O labels sit at their surface centers, which stay inside the anatomical
+// outline. The labels render outside the mirror group, so apply the same quadrant
+// mirror the group transform applies.
+const labelAnchors = computed(() => {
+  const centers = occlusalPaths.value.surfaceCenters
+  const mirror = (point: { x: number, y: number }) => ({
+    x: isLeftSide.value ? 50 - point.x : point.x,
+    y: isLower.value ? 50 - point.y : point.y
+  })
+  const anchor = (surface: string, fallback: { x: number, y: number }) =>
+    mirror(centers[surface] ?? fallback)
+  return {
+    M: anchor('M', { x: 12, y: 25 }),
+    D: anchor('D', { x: 38, y: 25 }),
+    V: anchor('V', { x: 25, y: 12 }),
+    L: anchor('L', { x: 25, y: 38 }),
+    O: anchor('O', { x: 25, y: 25 })
+  }
+})
 const toothName = computed(() => {
   const nameKey = getToothNameKey(props.toothNumber)
   const positionKeys = getToothPositionKeys(props.toothNumber)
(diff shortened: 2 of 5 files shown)


martinezsalmeron (changes requested):
Hi @Mr-Neutr0n, thanks a lot for picking this one up! 🙏 Moving from the uniform circle to per-category shapes is exactly the direction #48 asked for, and I like that you kept the 5-surface contract and the CSS-transform symmetry untouched, plus the vitest spec is a nice touch. I checked the branch out and ran it locally (chart + surface selector popup), and there are a few things to sort out before we can merge. All of them are small.

### Blocking

**1. `frontend-typecheck` is red.** The layer typecheck runs with `noUncheckedIndexedAccess`, so destructuring `Point[]` gives `Point | undefined` (34 errors in `ToothSVGPaths.ts`). Typing the corner lists as a 4-tuple fixes it cleanly, no runtime change:

```ts
type Quad = [Point, Point, Point, Point]

function buildOcclusalSurfaces(outer: Quad, inner: Quad) { ... }
function buildOcclusalDividers(outer: Quad, inner: Quad) { ... }
function buildOcclusalPaths(outline: string, outer: Quad, inner: Quad, ...) { ... }

const INCISOR_OUTER: Quad = [ ... ]   // same for the other 7 constants
```
I verified locally that this makes the errors go away.

**2. The `D` surface is a triangle, not a quad.** In `buildOcclusalSurfaces` the path is `M itr L obr L otr L itr Z`, which skips `ibr`. Visually the distal fill covers only half of the sector, and the missing half isn't clickable in the popup either (screenshot below, tooth 36 with D selected). The other four sectors are fine. It should be:

```ts
D: `M ${pointStr(itr)} L ${pointStr(otr)} L ${pointStr(obr)} L ${pointStr(ibr)} Z`
```

**3. The sectors don't follow the outline, so the outline disappears.** `OUTER` is always a rectangle, while the canine outline is a diamond and the premolar an oval. In `SurfaceSelectorPopup` the surfaces are filled (`--odontogram-fill`) on top of the outline, so the canine and premolar render as squares with the "anatomical" shape barely peeking out at the edges; on the chart the divider lines poke outside the diamond. The molar has the same issue in miniature (`OUTER` at 11/39 vs outline at 12/38). The lazy fix that works for all four shapes: wrap the surfaces `<g>` in a `<clipPath>` that uses the outline path (both in `ToothDualView.vue` and `SurfaceSelectorPopup.vue`), or alternatively bring the outer points inside the outline. Either is fine by me.

### Strongly suggested (UX)

**4. Molar fissures read as a "missing tooth".** The full-length `+` and `×` across the whole crown look almost identical to the missing-tooth overlay (which is a diagonal cross from 8,8 to 42,42). On the chart, 17/18/27/28 now look crossed out at a glance. Keeping the fissure lines inside the central `O` area (or dropping the diagonals and keeping only a short cross) would fix it. Note also that in the popup these highlight lines are drawn *before* the surfaces, so the filled sectors hide them entirely there.

### Housekeeping (project checklist)

- Add an entry under `## Unreleased` in `backend/app/modules/odontogram/CHANGELOG.md` (CLAUDE.md asks for it whenever a module is touched).
- Optional but welcome: a one-line assertion in the vitest spec that every surface path closes with 4 vertices would have caught (2).
- Small nit: the PR description says "No local test suite detected", but `cd frontend && npx vitest run tests/odontogram` runs your spec (it passes ✅).

(I have a before/after screenshot of the chart and the popup from my local run; I will attach it in a follow-up comment.)

Once (1), (2) and (3) are in I'm happy to merge; (4) I'd really like too, but tell me if you'd rather leave it for a follow-up. Thanks again for the contribution! 🦷

ZoliQua (changes requested):
Thanks for picking up all four points from the earlier round — the tuple types, the four-vertex D, the clip path and the short central cross are all in. One regression from the second commit blocks this, then two smaller things, and the branch now conflicts with `main` and needs a rebase (there is no CI run on the current head).

- `ToothDualView.vue` L973–981: the outline redrawn on top uses `class="tooth-occlusal"`, and the scoped rule at L1117 sets `fill: var(--odontogram-fill)`; a stylesheet rule wins over the `fill="none"` presentation attribute, so this path paints an opaque surface over the whole treatment-overlay group and the fissures. On the chart, occlusal fillings, caries and patterns disappear. Please give it its own class, e.g. `.tooth-occlusal-edge { fill: none; stroke: var(--odontogram-outline); }`, and re-check a tooth with an existing O filling plus a planned MOD.
- `ToothSVGPaths.ts` L496–498 (`INCISOR_OUTER`): the comment above says the outer quads may extend past the silhouette because callers clip, but the incisor quad (y 14–36) sits four units inside the outline (y 10–40), so the incisal and cervical bands are neither fillable nor clickable; premolar and molar have the same gap in miniature. Could you over-size the OUTER quads (incisor y 6–44, premolar y 4–46, molar 6–44) and add a spec asserting each quad encloses its outline's bounding box?
- `tests/odontogram/toothOcclusalPaths.test.ts` L2: other specs import layers through the `#module-layers/...` alias (see `tests/india_gst`), which also resolves inside Docker; please use it here too. And a quick look at 13 and 11 in the surface popup: the M/D/V/L labels are still at the circle-era positions and may overlap the new outlines.

ZoliQua (approved):
All three are in, and I checked the blocking one in the browser rather than in the diff.

- **The occlusal overlay regression is gone.** `.tooth-occlusal-edge { fill: none; stroke: var(--odontogram-outline) }` now owns the redrawn outline. I ran the chart on a dev server, picked a composite filling and applied it to the occlusal surface of a molar: the blue fill renders inside the crown with the fissure lines still visible around it. Existing work on the chart reads correctly too — the bowtie M+O fill on 17, the hatched planned work on 16 and 46.
- **The outer quads are over-sized** (incisor 6–44, canine 5–45, and so on) and `toothOcclusalPaths.test.ts` asserts each surface band encloses its outline's bounding box. Four specs, all green locally.
- **The spec imports through `#module-layers/…`**, matching `tests/india_gst`.
- The surface popup's M/D/V/L labels now sit inside their bands for each tooth shape — I opened it on 47 and they do not collide with the new outlines.

The branch also merges cleanly onto `main` now, and eslint on the three changed files is clean apart from the pre-existing `v-html` warning.

CI had never run on this head: the workflow run was sitting in "action required" since 11 September, waiting for a maintainer. I released it and it is running now. Nothing here touches a CODEOWNERS path, so once it comes back green I will merge this myself — no need to do anything further unless CI finds something.

One thing I saw that is **not** yours: on my dev server the console logs `Failed to resolve component: UDropdownMenuItem/Group/Separator` from `TreatmentBar.vue`, which this PR does not touch. I will check whether that reproduces on `main` and file it separately if it does.

ZoliQua:
@martinezsalmeron — gentle second ping on this one, since you cleared a lot of the queue on Sunday and this may have slipped past. Everything in your 2 September review is in the current head (I listed the five points with evidence above), CI is 12/12 green and it merges cleanly. It is blocked only by that review still standing as "changes requested". Happy to merge it the moment you clear it, or tell me if you would rather I dismiss the review on your behalf. @Mr-Neutr0n, sorry for the wait — nothing further is needed from you.

martinezsalmeron (approved):
Re-reviewed the current head (438a90f3) locally, merged onto today's main with no conflicts, and ran it in the browser: chart + surface popup on 11, 13, 14 and 36. All five points from my 2 September review are in, and the chart now reads clearly better than the circles — distinct incisor/canine/premolar/molar silhouettes, fills and clicks clipped to the outline, D is a full sector, molars no longer look crossed out. Vitest spec green, CI 12/12. Thanks @Mr-Neutr0n, and thanks @ZoliQua for the second round and the follow-up.

Issue #365, opened 2026-09-02
payments core: prerequisites for gateway modules (upi method, UI slots, idempotency key)


martinezsalmeron:
Prerequisite for #263 (India payments: UPI / PhonePe) and for the Stripe work @BabuBahir offered in the #210 thread. Both are payment-*gateway* modules that must extend `payments` without touching it (ADR 0010). Today `payments` gives them nothing to hook into, so this issue collects the small, provider-neutral core changes that should land **before** any gateway module starts. Keep it to exactly this list; anything provider-specific belongs in the gateway module.

## 1. `upi` payment method

- Add `"upi"` to `PAYMENT_METHODS` (`payments/models.py`) and to the `PaymentMethod` `Literal` (`payments/schemas.py`).
- i18n label `payments.methods.upi` in every locale we ship.
- Chip in `PaymentCreateModal` under the secondary methods, **gated on `clinic.country === 'IN'`** so non-India clinics don't see it. Follow the existing `SECONDARY_METHODS` pattern.
- `reports/by-method` and the dashboard `PaymentMethodsTile` pick it up automatically; verify, no new code expected.
- Optional, same PR if trivial: `"netbanking"`.

## 2. Extension slots in the `payments` UI

`payments` registers into other modules' slots but renders no `<ModuleSlot>` of its own. A gateway module needs these entry points; all render nothing when no module is registered:

| Slot | Where | `ctx` | Purpose |
|---|---|---|---|
| `payments.collect.actions` | next to the "New payment" button (payments page, `PatientPaymentsPanel`, `BudgetPaymentsCard` "Cobrar") | `{ patient?, budget?, prefer_invoice_id? }` | gateway module adds its own "Collect via …" action, opens its own panel |
| `payments.list.row.meta` | payments list row | `{ payment }` | provider badge, verification / settlement state |
| `payments.ledger.row.meta` | patient ledger row | `{ entry }` | same, on the patient ledger |
| `payments.detail.sections` | payment detail | `{ payment }` | provider ids, settlement, provider event log |

Register nothing from `payments` itself; document the slot names and `ctx` shapes in `payments/CLAUDE.md` under "Frontend slots rendered". The `condition` on the gateway side handles country gating, same as `verifactu` / `india_gst` (`slots.client.ts`).

## 3. Idempotency key on `record_payment` (optional, recommended)

[...]

PR #373, opened 2026-09-02
feat(payments): razorpay online payment integration

### LINKED ISSUE #263: India payments: UPI / PhonePe integration (separate module)
Requested in the #210 thread: UPI / PhonePe support for Indian clinics, now that the india_gst compliance module has shipped. @tresundios volunteered to explore it as a separate feature after the GST core; @BabuBahir offered to look at Stripe.

Scope notes before any code:
- Separate module(s), following the payments-module architecture (ADR 0010) — payment *methods/gateways* should extend `payments`, never billing.
- Design first (needs-design): gateway abstraction vs per-provider modules, reconciliation with the existing Payment + allocations model, and what "paid via UPI" means for the GST reconciliation report.
- A design sketch in this issue gets maintainer review before implementation — that saved a lot of rework on #210.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

BabuBahir:
A rough attempt to Fix #263

Added Razorpay gateway for online payment collection.

_Note 1 : RAZORPAY_KEY_ID and RAZORPAY_KEY_SECRET are to be added both in .env  and backend/app/config.py
 Note 2 : KYC of RAZORPAY is not required so Test Mode is enough for dummy Capture, Order , Payments_ 

Backend:
- New razorpay_service: lazy SDK client.
- POST /api/v1/payments/razorpay-order: create a Razorpay order server-side.
- POST /api/v1/payments/razorpay-verify: verify signature, then record the payment via the existing workflow with method='razorpay'.
- Add 'razorpay' to PaymentMethod / PAYMENT_METHODS.
- config: RAZORPAY_KEY_ID / RAZORPAY_KEY_SECRET settings.

Frontend:
- PaymentCreateModal: Pay with Razorpay flow — load checkout.js, create the order via the backend, open the checkout popup, and verify server-side on the handler callback.
- i18n locales: razorpay method label + button/error strings.

Env:
- .env.example, docker-compose.yml: Razorpay key passthrough. package: razorpay

 
Steps: 

1.  Add amount 
image

2.  From the Payment Options Choose NetBanking option ....(which will appear in a moment )
image

3.  We can simulate "Success" or "Failure"  ..
image

4. A paymentID is created which also reflects in RazorPay 
image

image

5. The refund option is also working..
image


BabuBahir:
Any comments on this @ZoliQua  @martinezsalmeron  @tresundios

martinezsalmeron (changes requested):
Thanks for taking this on — the shape of the server side is right (order created with the secret on the backend, HMAC checked with `compare_digest` before anything is written, lazy import so a backend without keys still boots), and touching all nine locales is appreciated. I went through the diff, the CI logs and the payments module contract; I also merged `main` into the branch locally to see what would actually land. I can't run the happy path (no Razorpay test account on my side), so the findings below are from code + CI, not from the checkout popup.

Verdict: **changes requested**. Two of the issues are architectural and the rest are correctness/security — I'd rather the next round be a small design sketch in #263 than another code iteration here.

## 1. Architecture — this belongs in a separate gateway module, not inside `payments`

#263 is labelled `needs-design` and says explicitly: *separate module(s), payment methods/gateways extend `payments`*, and *a design sketch gets maintainer review before implementation*. #370 (merged last week) built exactly the extension points for that: the `payments.collect.actions` slot (rendered next to "New payment", "Cobrar" and "Collect" in every host), `upi`/`netbanking` methods, and `idempotency_key` on `record_payment`. See `backend/app/modules/payments/CLAUDE.md` → "Frontend slots rendered": *"Entry points for payment-gateway modules (#365 / #263). `payments` registers nothing into them. Country gating is the gateway's `condition`."*

This PR instead hard-wires Razorpay into the core module: a second submit button inside `PaymentCreateModal.vue`, two routes in `payments/router.py`, schemas, `settings`, `pyproject`. Every clinic in every country now gets a "Pay with Razorpay" button that fails with a 400 unless the operator set INR keys.

What it should look like (same pattern as `india_gst` / `verifactu`):

- `backend/app/modules/razorpay/` with `manifest.depends = ["payments"]`, `installable=True`, `auto_install=False`, `removable=True`. Own router (`/api/v1/razorpay/order`, `/api/v1/razorpay/verify`), own permission (`razorpay.collect`), own CLAUDE.md/CHANGELOG/docs.
- A slot component registered into `payments.collect.actions` from `frontend/plugins/slots.client.ts`, with `condition` = clinic country `IN` (the `useClinicCountry` composable already exists) — the payments module and its modal stay untouched.
- It calls `record_payment` from `payments` (allowed, it's in `depends`) — no new code in `payments`.

## 2. Keys are per-deployment, but money is per-clinic

`RAZORPAY_KEY_ID/SECRET` live in global `settings`, so one Razorpay account per DentalPin instance. In a multi-clinic deployment every clinic's collections land in the same account. Keys should be per-clinic module settings (the way `whatsapp_webhook` / `integrations` store their tokens), which also gives you the "not configured → don't show the button" gate for free. This is a design decision for #263, so please put it in the sketch.

## 3. Amount is not verified — the client decides what gets recorded

The Razorpay signature is `HMAC(order_id|payment_id)`; it proves *a* payment happened for that order, it says nothing about the amount. `/razorpay-verify` then records `payload.amount` straight from the browser, and `/razorpay-order` accepts any amount without persisting anything that links order → clinic/patient/amount (the `patient_id` it receives is unused). So a 1 INR order can be recorded as a 50 000 INR payment, and there is no way to reconcile.

Fix: after the signature check, fetch the payment from Razorpay (`client.payment.fetch(payment_id)`), assert `status == "captured"` and `order_id` matches, and record `amount / 100` and `currency` **from Razorpay's response**, never from the request. Put `notes={"clinic_id": ..., "patient_id": ...}` on the order so the dashboard side is reconcilable too.

## 4. Replay — the same checkout response can be recorded N times

Nothing stops `POST /razorpay-verify` being sent twice with the same `(payment_id, order_id, signature)`: two `Payment` rows for one captured payment. `record_payment` already has `idempotency_key` (#365) — pass `idempotency_key=f"razorpay:{razorpay_payment_id}"` and this is a one-liner.

## 5. Currency mismatch

The frontend hard-codes `currency: 'INR'` for the order; the backend records the payment in `ctx.clinic.currency`. For a EUR clinic that's a 100 INR order recorded as 100 EUR. Use the clinic currency for the order (and refuse when it isn't one the gateway account supports). With #1 this becomes moot because the button only exists for IN clinics — but the backend should still not trust the client's currency.

## 6. `razorpay` is not a valid method — description says it was added, the diff doesn't

`PAYMENT_METHODS` (models.py) and the `PaymentMethod` Literal (schemas.py) are unchanged, and only `payments.new.razorpay*` labels were added, not `payments.methods.razorpay`. Consequences I could trace:

- Payments list / ledger rows render the raw key `payments.methods.razorpay` (`pages/payments/index.vue` uses `t(\`payments.methods.${p.method}\`)`).
- Refunding one of these payments fails with 422: `RefundConfirmModal` always offers the source payment's method, and `RefundCreate.method` is the Literal.
- The module's documented invariant ("methods are a closed list; models + schemas must stay in sync") is broken.

Suggestion: don't add a method at all. Razorpay's payment object carries `method` (`upi`, `netbanking`, `card`, `wallet`, `emi`); map it onto the existing closed list (`upi` / `netbanking` / `card` / `other`) and keep the gateway payment id in `reference`. That's also what the GST reconciliation in #263 wants ("what does *paid via UPI* mean").

## 7. Dependency isn't locked → the feature is dead in the Docker image

The branch predates #371: there is no `uv.lock` on it, and after merging `main` `uv lock --check` fails (I verified). The Dockerfile installs with `uv sync --frozen`, so the image would ship **without** `razorpay` and both endpoints would answer 400 "package not installed". `cd backend && uv lock` with uv 0.12.9 fixes CI.

Honest question whether the SDK is worth it at all: order creation is one `POST https://api.razorpay.com/v1/orders` with basic auth (httpx is already a dependency, and it's async), payment fetch is one GET, and the signature is stdlib `hmac`. The SDK is synchronous `requests`, so calling it inside an `async def` route blocks the event loop for the duration of the Razorpay round-trip. I'd drop it.

## 8. Frontend flow

- **Errors inside `handler` are swallowed.** The `try/catch` around `payWithRazorpay` doesn't cover the callback Razorpay invokes later. If `/razorpay-verify` fails (network blip, 400), the money is captured at Razorpay, nothing is recorded, and the user sees nothing — the modal just stays open. Worst case in a payments flow: silent orphaned capture. Needs a visible error including the `razorpay_payment_id` so staff can reconcile; a `payment.captured` webhook as the source of truth would be the robust version (`app.core.webhooks` from #347 has the signing helpers), but at minimum the callback must report failure.
- `isSubmitting` resets in `finally` before the popup even opens; no `modal.ondismiss`, so closing the popup silently returns to the form.
- The method chip the user selected is ignored when they click Razorpay, and the footer now has two competing submit buttons. As a slot action next to "New payment" this confusion disappears.
- Hard-coded English in the checkout: `'Clinic Payment'`, `'Medical Services Payment'`, and the theme colour. Use the clinic name + i18n.

## 9. CI (all three failures are in the diff)

- backend-lint: `I001` import order ×2 (`razorpay_service.py`, `router.py`), `N818` → `RazorpayNotConfiguredError`. Then `uv lock --check` (see #7).
- frontend-lint: 7 errors in `PaymentCreateModal.vue` — `no-explicit-any` ×3 (type `window.Razorpay` and the error instead), member-delimiter ×2, comma-dangle, trailing spaces.
- frontend-typecheck: `color="black"` (line 679) is not a Nuxt UI colour; `neutral` is.

## 10. Missing per the repo checklist (CLAUDE.md)

No tests (signature ok / bad / not-configured / replay / amount tamper), no `CHANGELOG.md` entry, no rows in `docs/technical/payments/permissions.md` for the two endpoints, no user-manual screen docs, no `.env.example` explanation of the INR-only nature. `.env.example` says *"leave empty to disable the online payment button"* but nothing implements that.

## Proposed next step

Rather than fixing all of this in place, post a short design sketch in #263 covering: module boundary (`razorpay` module depending on `payments`, slot + IN condition), per-clinic keys, server-side amount/status verification with idempotency, method mapping onto the closed list, and whether v1 includes the webhook. Once that's agreed the implementation is mostly moving the code you already have into the module and deleting the parts the slot makes unnecessary. Happy to review the sketch quickly.

ZoliQua (changes requested):
Thanks for turning the earlier review into a proper module — `backend/app/modules/razorpay/` is genuinely close to what #263 asked for: per-clinic Fernet-encrypted keys, `/verify` with no client `amount`, re-fetch + `captured` check, idempotency key, method mapped onto the closed list, slot registration through `payments.collect.actions`, own Alembic branch with a round-trip test, docs and tests in the right places. Most of what follows is about getting the branch to contain *only* that module.

1. **The old inline integration came back with the merges.** `payments/router.py` (`/razorpay-order`, `/razorpay-verify`), `payments/schemas.py`, `payments/razorpay_service.py`, `PaymentCreateModal.vue`, the nine `payments` locale edits, `app/config.py`, `docker-compose.yml` and `.env.example` are all still in the diff — your own `INSTRUCTION.md` says these should be back at `57f7c536`. `git checkout upstream/main -- backend/app/modules/payments backend/app/config.py docker-compose.yml .env.example` and the diff shrinks to the module. Keys belong in `razorpay_settings`, never in `Settings` (the `whatsapp_webhook` / `nav_online` pattern). That also removes every point from Ramón's earlier review that is still live through the old code (client-supplied amount, global secret, `method="razorpay"` outside the closed list).
2. **Checkout option name.** `RazorpayCollectButton.vue:103` and `types/razorpay.d.ts` use `key_id`; checkout.js requires `key`, so the popup will not open as is. While there: `modal: { ondismiss }` to reset `paying`, and pass the clinic name instead of `'DentalPin'`.
3. **`uv.lock`.** Unchanged on the branch; the image installs `--frozen`, so today the SDK is not in the image and `_client()` raises "package is not installed". `cd backend && uv lock` and commit. Also `client.order.create` / `payment.fetch` are synchronous `requests` calls inside `async def` — `await asyncio.to_thread(...)`, or two `httpx.AsyncClient` calls with basic auth (the signature is already stdlib).
4. **`is_active` is not enforced.** `create_order` and `capture_and_record` only check `settings is None`; toggling "Accept online payments" off still creates orders and records payments. Raise `RazorpayNotConfiguredError` when `not settings.is_active` in both paths, with a test.
5. **Secret rotation.** The settings page says "leave blank to keep the stored secret" but `RazorpaySettingsUpdate.key_secret` has `min_length=1` → 422. `str | None`, re-encrypt only when provided.
6. **Conflicts with `main`.** `alembic.ini` and `pyproject.toml` replace the `sms_gateway` entries instead of appending; after merging, re-run `node scripts/modules-json.mjs ../backend/app/modules --prefix /module_layers` and `python backend/scripts/generate_catalogs.py` (the catalog is reported stale).
7. **Locales.** The host now ships `ar.json`; please add `razorpay/frontend/i18n/locales/ar.json` (27 keys) and list it in the layer's `nuxt.config.ts`.
8. **Lint.** `ruff check` reports 15 (`I001`, `N818`, `UP037`, `W292` on every new file) and `ruff format` 11 files; eslint `eol-last` on every razorpay frontend file. `ruff check --fix && ruff format` and `npm run lint -- --fix` clear almost all of it.
9. **Housekeeping.** Delete `INSTRUCTION.md` and the `/.vs` gitignore line; manifest `author` → your handle (see `india_gst`); fill `last_verified_commit` in `docs/technical/razorpay/*.md`; one index on `razorpay_settings.clinic_id` is enough; the host's auto-imported `useClinicCountry()` replaces `useRazorpayCountry`; quantize `amount` to `0.01` before `* 100`; a `docs/user-manual/{en,es}/razorpay/screens/settings.md` for the Settings → Integrations page would match `sms_gateway`.
10. **Design (for @martinezsalmeron).** #263's agreed first PR is a provider-neutral `payment_gateways` (adapter contract, `PaymentRequest` state, webhook creating the `Payment` in-transaction, async refunds) with Razorpay as the first adapter, and @tresundios has said he is preparing that. This PR is a fine single-provider v1 but has no webhook — a failed `/verify` after capture leaves money unrecorded. Ramón should decide whether this lands, with the fixes above, as the Razorpay adapter seed, or is folded into the `payment_gateways` PR. `backend/alembic.ini` and `frontend/app/config/permissions.ts` need his review either way.

Issue #44, opened 2026-04-14
feat: GDPR Compliance Module (Data Subject Rights, Consent, Audit, Retention)


martinezsalmeron:
## Summary

Build a GDPR compliance module for DentalPin to help European dental clinics meet their data protection obligations. Health data is classified as **"special category data"** under GDPR Article 9, requiring enhanced protection measures.

**Non-compliance risk:** Fines up to **4% of annual revenue or €20 million** (whichever is higher).

### Data Subject Rights (Chapter 3)

| Right | Article | Implementation |
|-------|---------|----------------|
| **Access** | Art. 15 | Export all patient data |
| **Rectification** | Art. 16 | Edit personal data |
| **Erasure** | Art. 17 | Delete or anonymize data |
| **Portability** | Art. 20 | Export in machine-readable format |
| **Restriction** | Art. 18 | Pause processing |
| **Object** | Art. 21 | Opt-out of certain processing |

**Healthcare exception:** Right to erasure can be refused when retention is legally required (medical records). Must explain legal basis to patient.

### Consent Management

- Explicit, informed consent before processing
- Track: when given, what it covers, who gave it
- Easy withdrawal mechanism
- Separate consents for: treatment, marketing, research, sharing with third parties

### Data Retention

- GDPR: Keep data only as long as necessary
- Healthcare laws: Often require 5-15+ years retention (varies by country)
- Module must support configurable retention policies per data type
- Automatic flagging when retention period expires

### Breach Notification

- Notify supervisory authority within **72 hours**
- Notify affected individuals if high risk
- Document all breaches (even minor ones)

### Audit & Accountability

- Log all access to personal data
- Data Protection Impact Assessments (DPIAs)
- Processing activity records
- Demonstrate compliance on request

---

### 1. Data Subject Request (DSR) Portal

Backend endpoints + frontend UI for handling requests:

```python
# Endpoints
POST   /api/v1/gdpr/requests           # Patient submits request
GET    /api/v1/gdpr/requests           # List all requests (staff)
GET    /api/v1/gdpr/requests/{id}      # Request details
PUT    /api/v1/gdpr/requests/{id}      # Update status
POST   /api/v1/gdpr/requests/{id}/export   # Generate data export
POST   /api/v1/gdpr/requests/{id}/erase    # Execute erasure
```

**Request types:**
- `ACCESS` - Generate full data export
- `RECTIFICATION` - Flag data for review/correction
[...]

PR #374, opened 2026-09-03
feat(gdpr): GDPR compliance module (data-subject rights, consents, retention, erasure, breach register - issue #44)

lamanji:
## Summary

- GDPR (EU 2016/679) compliance module (v1 backend base): data-subject requests with 30-day SLA (Art. 15-20, ticket tracker), append-only patient consents (Art. 7-8), retention policies gating erasure (Art. 5(1)(e)), partial erasure with immutable audit log (Art. 17), breach register (Art. 33-34), portability export (Art. 20).
- Maintainer review round addressed: closed erasure vocabulary, patient-first 404, anchor-date retention expiry, append-only consent trail, DSR DELETE dropped.

## Tasks done

- `feat(gdpr): GDPR compliance module` (data-subject rights, consents, retention, erasure, breach register)
- `Merge remote-tracking branch 'upstream/main'` (module-list conflicts in `backend/alembic.ini` + `backend/pyproject.toml`, resolved as union)
- `fix(gdpr): address maintainer review` (closed `ErasureCategory` Literal, 404-before-audit, `updated_at`-anchored expiry, append-only consents, DSR DELETE dropped, `now(UTC)`, dead code removed, v1 scope docs)

## Modules touched

- `gdpr`

## Checklist

- [x] `ruff check . && ruff format --check .` green
- [x] Tests added/updated with behaviour change (13 default incl. 5 new regression tests + uninstall roundtrip)
- [x] No cross-module imports outside `manifest.depends` (`patients` only)
- [x] Entry point registered in `backend/pyproject.toml`
- [x] `backend/app/modules/gdpr/CLAUDE.md` + `CHANGELOG.md` written/updated
- [x] Migrations on own branch (`branch_labels=("gdpr",)`)
- [x] Catalogs regenerated (`docs/modules-catalog.md`, `docs/events-catalog.md`)
- [x] Uninstall roundtrip test added and green
- [x] New events in `EventType` + documented in `docs/technical/gdpr/events.md`
- [x] New permissions gated with `require_permission` + listed in `docs/technical/gdpr/permissions.md`

## Test plan

- `scripts/dev-verify.sh --backend-only --with-pytest --module gdpr` (compose `db` up): backend-test 13 PASS + roundtrip 1 PASS (container)
- `scripts/dev-verify.sh --backend-only`: lint, manifest, catalog-freshness, docs-coverage, docs-layout all PASS
- End-to-end in the running stack: create a DSR, grant/withdraw/re-grant a consent (3-row trail), add a retention policy, run a partial erasure, check the audit log, report a breach, export a patient at `/api/v1/gdpr/*`
- Repro checks for the review fixes: `POST /erasure` with `clinical` returns 422; erasure of an unknown UUID returns 404 with zero audit rows

## Refs

- Part of #44 (v1: backend DSR/consent/retention/erasure/breach base; UI, retention-expiry flagging and access logging stay tracked in #44)

## Review

Tagging `@martinezsalmeron` for review/merge when convenient.


martinezsalmeron (changes requested):
Thank you for this one — GDPR has been sitting in the backlog since #44 and this is a solid base to build on. I went through it in depth on top of current `main`: module isolation, tenancy filtering, the `gdpr` Alembic branch, the tools contract and the docs are all in order, and the 8 unit tests plus the uninstall roundtrip pass locally (conflicts with `main` are just the two module lists in `pyproject.toml` and `alembic.ini`). 🙏

Because this module's whole job is to be *truthful* about what happened to a patient's data, I'd like to fix a few things before merging. I reproduced 1, 2 and 4 with small probe tests.

### Blockers

1. **The erasure audit log can report an erasure that never happened.** `ErasureService.execute` only knows `email` / `phone` / `identity` (via `category_to_fields`), but any category with a 0-year policy is reported as erased. Repro: policy `clinical` / 0 years → `POST /erasure {"categories": ["clinical"]}` → `erased_categories=["clinical"]`, an audit row is written with `fields_blanked=None`, and the patient is untouched. The tool description even suggests `clinical, billing, radiology`. Suggestion: make `categories` a `Literal["email", "phone", "identity"]` in the schema and the tool args so unknown values are a 422, and align the retention `data_category` vocabulary with it (or document the mapping explicitly).
2. **Erasure for a non-existent (or other-clinic) patient ends in a 500.** The audit row is written even when the patient lookup returns `None`, so the FK on `gdpr_erasure_audit_logs.patient_id` fails. Returning `None` from the service (404 in the router, error in the tool) before touching the audit table would do it.
3. **`retention_years` currently behaves like a boolean.** `erasure_eligible` only checks `retention_years == 0`; there's no anchor date, so a 15-year policy blocks forever and a 0-year policy is effectively an allowlist. #44 asks for flagging when a retention period expires. Two honest options: compute expiry against a real anchor (patient `updated_at` at minimum, last visit once a provider exists), or rename the field/docs to reflect what it does today. Genuinely open to your reasoning here.
4. **Consent history is overwritten on re-grant.** grant → withdraw → grant leaves a single row with `withdrawn_at=None`, so the withdrawal is gone — contrary to the docstring ("the row stays for audit") and to the consent timeline in #44. Appending one row per event (latest per purpose is a one-liner) would keep the trail intact.
5. **`DELETE /requests/{id}` hard-deletes a data-subject request**, and receptionists have that permission. DSRs are accountability records (Art. 5(2)); I'd drop the endpoint or turn it into a soft state.

### Nice to fix while you're there

6. `datetime.now()` is naive throughout; every other service in the repo uses `datetime.now(UTC)`. It only works because the container runs in UTC.
7. `ExportService.record_export` is unused and publishes `gdpr.request.created` with `request_id=None`, which breaks the payload contract documented for that event. Safe to delete.
8. Small leftovers: `clinic_id_of()`, `getattr(ctx, "user_id", None)` (`ClinicContext` always has `user_id`), `definition_year` in `ExportResponse`.
9. `request_type` has no `objection` (Art. 21) although the title mentions 15-21, and `restrict` / `rectification` don't affect the patient record. Fine as a ticket tracker for v1 — just worth stating in the docs.

### Scope vs #44

10. There's no frontend layer, so staff can only reach this through the copilot tools or raw HTTP. Could you change `Closes #44` to `Part of #44`? UI, retention-expiry flagging and access logging would then stay tracked there, the same way `suppliers` was the backend base of the procurement suite.
11. The export covers identity + consents + DSRs, not clinical / appointments / billing. Totally acceptable for a v1 if the docs say so.

Happy to re-review as soon as you push — the two erasure repros take me a minute to re-run. Thanks again for the work on this. 🙌

martinezsalmeron (approved):
Re-verified on top of current main: 13/13 unit tests + alembic roundtrip green, ruff clean, and my two HTTP repros now behave (`clinical` → 422, unknown patient → 404 with zero audit rows, no DELETE route). All five blockers closed. Thanks for the fast turnaround 🙌

Issue #230, opened 2026-08-21
Advanced statistics: outstanding invoices, demographics, inventory usage, prescriptions, operational KPIs (design first)


martinezsalmeron:
Roadmap phase 16. Five new statistic families for reports: financial (outstanding invoices), patients (demographics, new vs returning, visit frequency), inventory (most-used, cost, low-stock runway), medications (most-prescribed, trends), operational (completion rate, staff productivity).

**Design discussion required before any code.** As described it reads directly from 7 modules (billing, payments, patients, agenda, inventory, documents, treatment_plan), which breaks module isolation unless every one is in `depends` — and that makes the module nearly non-removable. Options to weigh here:
- Extend the existing `reports` module vs. a new module
- Declared `depends` vs. event-fed materialized stats tables
- Which families ship v1

⚠️ Constraint: no paid-vs-invoiced diff metrics — reports must not cross the payment and invoice axes.

lamanji:
**Proposal — extend `reports` (v0.2.0) as the analytics read-aggregator; ship remote families as owner-owned endpoints; no event-fed tables in v1.**

**1. Extend `reports` vs. a new module → extend `reports`.**
`reports` is already the designed cross-module read side: official, `removable=False`, no models, no events, computed on demand. A sibling `statistics` module wouldn't fix isolation by itself (it would need the same `depends`) and would split permissions, docs, tools, and the dashboard machinery in two. Keep one home: per-family permissions, `/reports/*` endpoints, one `tools.py` (agent-readable), one widget registry.

**2. `depends` vs. event-fed tables → today: neither extreme. Hybrid:**
- **Families whose sources are already inside `reports.depends`** (billing, payments, patients, agenda) stay **computed on demand** in `reports` — no new tables, no backfill, no event coverage risk. This covers financial, patients, and operational.
- **Families that live in removable modules** (inventory, documents) get their stat endpoints in the **owning module** and surface through the existing `reports.dashboard.widgets` slot (shared date-range `Ref`, same pattern `notifications`' settings pages already use). `reports.depends` does **not** grow to include them, so `inventory` and `documents` stay removable — this is the actual fix for the "nearly non-removable" concern, cheaper than either a mega-`depends` or a new module.
[...]

PR #385, opened 2026-09-05
feat(reports): financial family - aging buckets + issued trend, CSV, guard test (roadmap #230)

lamanji:
## Overview

Financial reporting family for `reports` (roadmap #230, first of two
parts; demographics/visits/productivity follow after this merges):
outstanding aging buckets + issued trend, strictly on the invoice axis
(never payments — an off-books guard test pins this structurally), CSV
export on list endpoints, and the `financial_report` copilot tool. The
`/reports/billing` page now reads the new endpoints instead of the
payments subsurface.

## Tasks done

- `feat(reports): financial family` — `services/financial.py`, routes,
  tool, card rewire, guard test.
- `docs(reports): describe financial family` — overview section.
- `test(reports): move financial + offbooks-guard tests` — collected by
  the `--module` scope.
- `fix(reports): expose reports.financial.read` in the frontend
  permissions mirror.

## Modules touched

- `reports`

## Checklist

- [x] `scripts/dev-verify.sh --fast` green (7 pass; only the known MSYS
  `modules-json-freshness` path flake; CI is the authority)
- [x] reviewer pass 0 blockers (first-party per L31)

## Test plan

- `scripts/dev-verify.sh --with-pytest --module reports`: 3/3 green
  (aging, trend, off-books guard) on a bare schema.
- `scripts/dev-verify.sh --fast`: 7 passed, 2 skipped, 1 MSYS-only fail.
- End-to-end: open `/reports/billing`, confirm aging buckets + trend
  render from invoice data with no settlement columns.

## Part of

- Part of #230 (patient-stats + operational families follow after this merges)

## Review

`@martinezsalmeron` for review/merge when convenient. Follow-up
already scoped: trend chart has no page section yet.


martinezsalmeron (changes requested):
Thanks a lot for this one — the service layer, the tests and the docs plumbing are in good shape (clinic-scoped queries on the existing `clinic_id+status` / `clinic_id+issue_date` indexes, `Decimal` throughout, CSV as a plain `?format=csv` exactly as agreed in #230, and the CLAUDE.md deps drift fixed). I exercised both endpoints against the demo clinic and the numbers match the DB. Two points I'd like to sort out before merging, plus one small tightening.

**1. The dashboard card quietly changes meaning**

`AgingReceivablesCard` ("Cuentas por cobrar", gated by `payments.reports.read`) used to read `/payments/reports/aging-receivables`, i.e. `earned − paid` per patient. That pairing (paid ↔ earned) is the *allowed* one under the off-books rule, so the card was fine as it was. After the rewire it shows gross `Invoice.total` for `issued|partial` invoices bucketed by due date. On the demo clinic:

- before: 742,50 (7 patients) · after: 545,00 (2 invoices)
- `FAC-2026-0002` (135,00 total, 67,50 already paid, due **tomorrow**) shows as 135,00 in "0-30 días" — not overdue yet, and overstated.

`/reports/payments` still renders the old definition, so we'd end up with two "aging" widgets that disagree. #230 asked to consolidate, but under the same title this reads as a redefinition. Also `PERMISSIONS.reports.financialRead` is added but never used — the card gate stays on payments.

Two ways forward, I lean to A:
- **A.** Revert the composable rewire (the receivables card is legitimate) and surface the new invoice-axis aging + issued trend as sections on `/reports/billing`, gated by `reports.financial.read`, labelled as invoice aging, with a "no vencidas" bucket instead of folding not-yet-due invoices into 0-30.
- **B.** Keep the rewire but rename the card, move the gate to `financialRead`, add the "no vencidas" bucket, and drop the aging card from `/reports/payments` so there is a single definition.

**2. The user manual describes sections that don't exist yet**

`docs/user-manual/{en,es}/reports/screens/reports_billing.md` document "Aging" and "Issued trend" on `/reports/billing`, but the page has neither (the only consumer is `pages/reports/index.vue`, and the trend has no UI yet, as you noted). Once (1) lands the docs will be true, so this mostly resolves itself — just flagging so they move together.

**3. Guard test (small)**

The token scan over `tools.py` forced rewording the module docstring to "single-axis difference" just to avoid the word `collected`. Scoping the tokens to identifiers (`InvoicePayment`, `allocated`, `refunded`, `balance_due`) or scanning only `services/financial.py` would let the prose go back to plain language.

Minor, no need to block on any of these: no `response_model` on the two new routes (OpenAPI shows nothing), `format` shadows the builtin, no HTTP-level test for the `reports.financial.read` gate, `_invoice()` creates a `User` per row. The branch isn't rebased on main but merges cleanly.

Happy to re-review as soon as you push — the family itself is the right shape.

PR #394, opened 2026-09-06
feat(auth): HttpOnly session cookies, rotating refresh tokens with reuse detection, CSRF (#353)

### LINKED ISSUE #353: feat(auth): HttpOnly cookie auth + refresh token rotation with invalidation
From an external implementer's audit of 0b47934b — the most substantive security finding. Needs an ADR before code.

**Current state (verified):**
- `frontend/app/composables/useAuth.ts:26-34` stores both tokens in non-HttpOnly cookies (`secure` in prod, `SameSite=Lax`, `maxAge` 7 days on both) and JS reads them to build the `Authorization: Bearer` header in `useApi`.
- The refresh token is a stateless JWT. `/refresh` mints a new pair but the old refresh token stays valid until expiry unless `token_version` is bumped — no replay defense for a stolen token.

**Risk:** any XSS yields a 7-day session on an app holding health data. Compounded by the current lack of CSP (separate issue).

**Direction to design:**
- Backend sets tokens as `HttpOnly + Secure + SameSite` cookies (`Set-Cookie` on `/login`, `/refresh`, `/logout`); frontend stops reading tokens from JS.
- Refresh rotation with server-side invalidation of the superseded token (per-token id or per-user chain), not just `token_version`.
- Decide CSRF posture once auth moves to cookies (SameSite=Lax + custom header, or token).
- Touches `useApi` refresh flow, SSE auth for the copilot stream, and the e2e fixtures that pin cookies.

### LINKED ISSUE #421: auth: grace window for concurrent refresh from two tabs (ADR 0023 follow-up)
Follow-up from the review of #394.

**Problem.** Two tabs whose access cookie expired in the same instant both call `POST /auth/refresh` with the same `dp_refresh`. The first rotates it; the second presents the now-revoked token, reuse detection fires and revokes the whole family, logging both tabs out. The client-side dedupe in `useAuth.refresh` only covers one tab.

**Standard mitigation.** Accept a token revoked less than N seconds ago (e.g. 30 s) whose `replaced_by` is the current live token of the family, without burning the family: answer with the live token's session again (or a fresh rotation of it) instead of 401. Anything older, or a revoked token whose successor was itself already replaced, keeps burning the family as today.

**Where.** `backend/app/core/auth/service.py::rotate_refresh_token` (the `row.revoked_at is not None` branch), a test in `backend/tests/test_auth_cookies.py` with two sequential refreshes of the same token inside the window and one outside it.

ZoliQua:
Implements **ADR 0023** exactly as merged in #381 (status flipped to *accepted* in this diff). Heads-up was posted on #353 before starting.

## Backend
- **Cookies set by the backend** on `/login`, `/setup`, `/set-password` and `/refresh`: `dp_access` (HttpOnly, path `/`, 15 min), `dp_refresh` (HttpOnly, **path-scoped to `/api/v1/auth/refresh`** — it never rides along anywhere else), `dp_csrf` (readable, the double-submit half). `Secure` in production, `SameSite=Lax`. Helper: `app/core/auth/cookies.py`.
- **Transition release**: the JSON body still returns the token pair, and `get_auth_token` accepts `Authorization: Bearer` **or** the cookie (header wins) — Zapier `dp_` tokens, scripts and the e2e API context are untouched.
- **CSRF**: cookie-authenticated unsafe methods must echo `dp_csrf` in `X-CSRF-Token` (constant-time compare) → 403 otherwise. Bearer requests skip the check by construction; public/webhook endpoints never carry the session cookie.
- **Rotation with reuse detection**: `auth_refresh_tokens` (migration `0007`; `jti` + `family_id`, issued/expires/revoked/replaced_by, UA hash, IP). `/refresh` revokes the presented token and issues its successor in the same family; presenting an already-revoked token **burns the family**. Legacy stateless tokens (pre-0023) are honoured once and migrated into a tracked family. `token_version` stays the log-out-everywhere hammer; `revoke_all_for_user` is there for the admin "sign out all" follow-up.
- **`POST /auth/logout`** (new): revokes the family via the access token's `fam` claim (the refresh cookie can't reach this route by design), clears the cookies, always 204. Deliberately lenient on CSRF — a forced sign-out isn't worth protecting; the gate is proven on a real mutating endpoint in the tests.

## Frontend
- `useAuth`: no token cookies any more — `isAuthenticated` = user loaded, the readable `dp_csrf` is the "session probably exists" hint (anonymous visitors cost no `/me`). Login/refresh/logout go through the cookies. **SSR**: the incoming `cookie` header is forwarded to the backend, and a server-side refresh relays the rotated `Set-Cookie` headers onto the Nuxt response (otherwise the browser would keep the revoked refresh token and burn its own family on the next refresh — the one subtle trap in this design).
- `useApi`: `credentials: 'include'` + CSRF header on unsafe methods; new `raw()` for blob downloads. New host composable `useSessionRequest()` gives module code the same options.
- The **20 call sites** that built a bearer header by hand (PDF/CSV/blob downloads in billing, budget, media, documents, purchase_orders, recalls, accounting_export, india_gst, clinical_notes; media uploads/deletes/updates; the copilot SSE `fetch`) migrated; `grep -rn accessToken frontend/app backend/app/modules` is empty.
- e2e fixture: stops pinning `access_token` by hand — the API login's `Set-Cookie` lands in the shared browser-context jar (cookies are host-scoped, so `:8000` → `:3000` on `localhost` just works).

## Tests
`tests/test_auth_cookies.py`: HttpOnly/path flags + body-transition, cookie-only auth on `/me`, CSRF 403 → 200 on a real PATCH, bearer path unchanged + CSRF-exempt, rotation + reuse burns the family (old *and* new token dead, rows revoked), cookie-only refresh, logout revokes + clears, legacy token migration. Existing auth / setup / copilot / integrations-public-API / module-isolation suites green (51). ruff clean, eslint clean, full `nuxt typecheck` clean, migration 0007 up → down → up verified.

## Follow-ups (not here)
Drop the token pair from the JSON body next release; admin "sign out all sessions" UI on top of `revoke_all_for_user`; nightly prune of expired rows.

Closes #353.

Closes #421 (grace window implemented in 73db83b4).


martinezsalmeron (changes requested):
Thanks — this is a careful implementation of ADR 0023 and the backend half is in good shape: rotation + reuse detection, the `fam` claim on the access token so `/logout` works despite the path-scoped refresh cookie, the limiter reading the cookie, `/setup` and `/set-password` going through the same `_start_session`, and a test file that actually exercises the CSRF gate on a real mutating endpoint. I ran it locally (branch + `main` merged, `ACCESS_TOKEN_EXPIRE_MINUTES=1` to force expiry) and drove it in Chrome: login, POST with the CSRF header, 401 → refresh → retry with the rotated `dp_csrf`, blob download via `credentials: 'include'`, logout (204, family revoked, cookies cleared) all work on the same-host topology.

Two things block the merge, both about where the cookies live rather than what they contain.

### 1. Blocker — split-host deployments lose the session entirely

`docker-compose.coolify.yml` (what runs demo.dentalpin.com) serves the app from `demo.dentalpin.com` and the API from `api-demo.dentalpin.com` (`ALLOWED_ORIGINS: ${FRONTEND_URL}` is there precisely because it is cross-origin). The three cookies are set without a `Domain` attribute, so they are host-only cookies of the **API** host:

- `document.cookie` on the app host never contains `dp_csrf` → `useSessionRequest().hasSession` is always `false` on both client and SSR.
- `init()` therefore never calls `/me` → every hard reload / deep link 302s to `/login`, and the client does not recover either.
- `refresh()` bails on `!hasSession` → the first 401 after 15 min logs the user out.
- `csrfHeaders()` has no value to echo → every cookie-authenticated POST/PUT/PATCH/DELETE is a 403 (Lax cookies *are* sent on the same-site XHR, so the request authenticates and then fails the CSRF check).

Net effect: login works, and everything after it breaks on the first mutation or reload. Since `main` auto-deploys to the demo, this cannot merge as is.

Suggested fix (small): an optional `COOKIE_DOMAIN` setting (e.g. `.dentalpin.com`) passed as `domain=` to the three `set_cookie` / `delete_cookie` calls in `cookies.py`, empty by default (host-only, as now, which is what e2e and single-host `docker-compose.prod.yml` need), and documented in `docker-compose.coolify.yml` next to `FRONTEND_URL`. The refresh cookie stays path-scoped; a `Domain` cookie is still only sent to `/api/v1/auth/refresh`, which no Nuxt route has.

### 2. Blocker — SSR refresh can never succeed, so reload after expiry loses the page

`dp_refresh` is path-scoped to `/api/v1/auth/refresh`, so the browser never sends it with a page request. The server-side `refresh()` forwards the incoming `cookie` header, which by construction has no `dp_refresh` → the backend answers 401 (`presented` is `None`). Reproduced with curl and in the browser:

```
# only dp_csrf reaches Nuxt once dp_access has expired
curl -s -o /dev/null -w "%{http_code} -> %{redirect_url}\n" http://localhost:3000/patients -H "Cookie: dp_csrf=…"
302 -> http://localhost:3000/login
```

In Chrome, reloading `/patients/<id>` after expiry: SSR redirects to `/login`, the client middleware refreshes successfully on hydration and then `navigateTo('/')` — the user lands on the dashboard with a login-page flash and the deep link is gone. Today on `main` the same reload works because the refresh cookie was readable at `/`. Opening a patient in a new tab after 15 idle minutes is a very common clinic gesture, so this is a regression rather than a nit.

Two ways out, your call:

- **(a)** set `dp_refresh` with `path="/"` (HttpOnly makes the path scoping mostly cosmetic) so the SSR relay you already wrote becomes live — **but** then the SSR cookie forwarding must become per-request: after the middleware refreshes, later `useApi` calls in the *same* render still forward the original `cookie` header (no `dp_access`, the now-revoked `dp_refresh`) → second refresh → reuse detection burns the family. Storing the merged header on `useRequestEvent().context` and reading it in `useApi`/`useAuth` instead of `useRequestHeaders` fixes that in a few lines; `_mergeCookieHeader` already does the merging.
- **(b)** keep the scoping, drop the dead SSR relay, and make the server-side path a no-op when `hasSession && !user`: skip the redirect, let the client refresh on hydration. Needs care with anything awaited on the server (`layouts/default.vue` awaits `ensureLoaded()`), which would now 401 during SSR.

Either way an e2e test "reload a deep link after the access cookie expired lands on that URL, no `/login` in between" would have caught this and is cheap with `ACCESS_TOKEN_EXPIRE_MINUTES` low in the e2e env.

### Should fix (small)

- **CSRF token rotates on every refresh.** It works in Chrome only because Nuxt's `useCookie` subscribes to the CookieStore API; on browsers without it (older Firefox/Safari) a `useSessionRequest()` created before the refresh keeps the stale `dp_csrf` → the retry in `useApi` and every later mutation from an already-mounted component get 403, and a second tab breaks as soon as the first one refreshes. Cheapest fix on the backend: in `/refresh`, `csrf_token=request.cookies.get(CSRF_COOKIE) or new_csrf_token()` — the token is per family, not per access token.
- **"Legacy stateless token is honoured once"** (docstring, test name, PR text) is not what the code does: nothing marks a `jti`-less token as used, so it can be replayed until its own `exp` (7 days). Acceptable for a transition release, but say so, or reject legacy tokens whose `iat` is older than the deploy.
- **Two tabs refreshing in the same instant** — the second presents the just-revoked token and burns the family for both. Low probability, but a short grace window (accept a token revoked < N seconds ago whose `replaced_by` is the current live token, without burning) is the standard mitigation; fine as a follow-up issue.

### Nits

- `dependencies.py`: the `# noqa: E402` on the `.cookies` import is not needed.
- `revoke_all_for_user` is unused — fine as the hook for the follow-up, just noting.
- Unrelated, spotted while testing: `GET /billing/invoices/{id}/pdf` validates `locale` against `^(es|en|ta)$`, so a UI in pt/pl/it/de/ar gets a 422 on "Download PDF". Pre-existing, not this PR.

Requesting changes for 1 and 2; happy to re-test in the browser once they land.

martinezsalmeron (approved):
Re-reviewed after the fix-up commits, with `main` merged locally and driven in Chrome (`ACCESS_TOKEN_EXPIRE_MINUTES=1`): login, cookie-authenticated PATCH through the CSRF gate, client-side 401 → refresh → retry, **SSR refresh on a deep link after expiry** (200 with the rotated jar, no `/login` detour — both with curl dropping `dp_access` and with a real idle browser), logout (204, family revoked, cookies cleared). Both previous blockers are fixed; the `COOKIE_DOMAIN` test and the `session-expiry` e2e cover them.

Three things I found and pushed on top of the branch (73db83b4), so nothing to do on your side:

1. **Family burned by an SSR error re-render.** After the access cookie expired, opening an unknown route (`/billing/invoices` — Nuxt 404) made the SSR refresh succeed and then Nuxt's error page render presented the *pre-rotation* cookie again → reuse detection revoked the family → logged out on the next expiry. Same shape as the two-tab race in #421, so I implemented the grace window you described there in `rotate_refresh_token` (`REFRESH_REUSE_GRACE_SECONDS`, default 30, 0 disables): a token revoked inside the window whose successor is still live gets that successor back; anything older still burns. Verified in the browser with two tabs reloading at once after expiry — both stay in, no revoked row. Test added; the burn test pins the grace to 0.
2. **`_refresh_rate_key` never saw the cookie.** The browser refresh has no body, so `request.json()` raised before the `or request.cookies.get(...)` line ran — every browser refresh was keyed by IP, which behind the proxy is one bucket per tenant (exactly what the docstring says it avoids). Cookie read first now; unit test with an empty-body request.
3. **Logout after the access cookie had expired revoked nothing.** The browser drops `dp_access` at max-age, so `/logout` had no `fam` candidate and only cleared cookies. `dp_refresh` is on `path=/` now, so it is a candidate too; test added.

Not in this PR (pre-existing, filed separately): raw `fetch` blob downloads (invoice/budget/PO PDFs, CSV exports) don't go through the 401 → refresh path, so clicking "Download PDF" after 15 min idle shows "Not authenticated". `useApi().raw()` is the right home for that.

Issue #322, opened 2026-08-27
perf(frontend): build memory budget — i18n (9 locales × 34 layers) is the superlinear multiplier


martinezsalmeron:
From the resource audit (see PR #321 for the immediate dev-compose fix).

## Facts

- The prod Nitro build OOMed at 4GB and is now pinned at 6144MB (`frontend/Dockerfile.prod:21-22`, commit 7b171bf1; same in `ci.yml` — "#304 tipped it over"). Every new module layer and every new locale pushes that ceiling further.
- Translation surface today: **9 locales × (2,488 host keys + 1,391 layer keys) ≈ 35k strings, ~2.3MB across 181 JSON files**.
- The **host** locale files already carry module namespaces (`budget` 301 keys, `invoice` 263, `odontogram` 180, `treatmentPlans` 176, `reports` 161… in `frontend/i18n/locales/en.json`, 113KB; `ta.json` 236KB). So host i18n grows with module count too, and those files are loaded eagerly for every user — plus the full `en` fallback set on top for non-English users (`i18n.config.ts` `fallbackLocale: 'en'`).

## Proposals

1. **Move module namespaces out of the host locale files** into the owning module's layer `locales/` (the mechanism already exists and lazy-loads per locale). Shrinks the always-loaded host bundle and makes each module carry its own i18n weight.
2. **CI memory gate**: record peak RSS of the prod frontend build (e.g. `/usr/bin/time -v` in the release/e2e build job) and fail/warn on regression. Right now the only signal that a module or locale tipped the build over is a production OOM.
3. Revisit whether the `en` fallback needs to load the *full* English set per layer, or only missing-key namespaces.

Related: whether prod should keep baking all 34 layers is the architecture discussion (linked separately).

ZoliQua:
Measured proposal 1 before shipping it, and the numbers argue against it in its current form — **moving the host namespaces into the layers makes the build-memory problem worse, not better**:

| Run (same machine, macOS, default Node heap, prod build) | Peak RSS | Wall |
|---|---|---|
| main as-is (24 layer-locale sets + host) | **3.90 GB** | 61 s |
| after moving 1,900 host keys into 9 new layer locale dirs (34 sets) | **OOM** (`Ineffective mark-compacts near heap limit` past ~4.4 GB) | — |

[...]

PR #406, opened 2026-09-07
perf(i18n): one locale chunk per language instead of one per layer in prod builds (#322)

ZoliQua:
Follow-up to the measurement on #322. A local Nuxt module (`frontend/modules/i18n-consolidate.ts`, installed before `@nuxtjs/i18n` because Nuxt loads `modules/` first) merges every layer's `i18n/locales/<code>.json` into one generated file per locale under `node_modules/.cache/dentalpin-i18n/`, prepends it to the host locale entry, and drops the per-layer i18n config. Precedence is unchanged (earlier layer wins over later layer, host wins over all; `mergeLayerLocales` is exported and covered by `tests/i18n-consolidate.test.ts`). Dev mode is untouched so editing a layer locale still hot-reloads; `NUXT_I18N_CONSOLIDATE=0` opts out of a prod build.

What it buys, measured on the full 34-layer prod build (same machine, `--max-old-space-size=6144`, `/usr/bin/time -l`):

| | main | this PR |
|---|---|---|
| client chunks in `.output/public/_nuxt` | 531 | 326 |
| client bundle size | 7.94 MB | 7.60 MB |
| locale files the bundler processes | 214 layer + 10 host | 9 merged + 10 host |
| requests to load one locale at runtime | 1 host + up to 26 layer chunks | 2 |
| build wall time | 63–67 s | 58–59 s |
| peak RSS | 5.6–5.7 GB | 5.8–5.9 GB |

**It does not lower build memory.** That is the important finding for #322 and it corrects my August comment: the file count of the i18n layer locales is not the multiplier. Two more data points from the same session: with the default Node heap (4.3 GB on Node 25) main *and* this branch both OOM today (main fit in 3.9 GB on 29 August, so the ceiling moved with the layers merged since), and the peak grows with the layer count itself: 0 layers 4.08 GB, 13 layers 5.02 GB, 34 layers 5.6 GB in the same metric, i.e. roughly 50–60 MB per layer regardless of locales. Details in the issue comment.

So this is a chunk-count/runtime win with a smaller output and a slightly faster build, not a memory fix; if the extra moving part is not worth that, closing it is fine and the numbers still stand on the issue.


martinezsalmeron (approved):
Reviewed and verified locally on top of current main (40 baked layers, 28 with i18n):

| | consolidation off | this PR |
|---|---|---|
| client chunks in `.output/public/_nuxt` | 593 | 335 |
| client bundle | 8.1 MB | 7.6 MB |
| build wall time | 48 s | 43 s |
| peak RSS | 5.50 GB | 5.49 GB |

- The 10 generated files are byte-equal to an independent deep merge of the 268 layer files in i18n's own precedence order; there are currently zero cross-layer key conflicts, and the 7 host/layer overlaps keep the host value as before.
- Browser check on the prod build: es / pl / ar (RTL) render layer strings correctly on dashboard and `/recalls`, no console errors.
- One comment fix pushed: in Nuxt 4.4 `resolveModules` registers the `modules` array before `modules/`, so `@nuxtjs/i18n` is actually installed first. It does not matter because i18n only reads layer configs in its `modules:done` hook, but the header said otherwise.

Agree with the conclusion on #322: chunk-count/runtime win, not a memory fix. Merging.

Issue #46, opened 2026-04-21
Refactor permission system: move permissions from code to database with per-clinic customizable roles


martinezsalmeron:
## Context

Currently permissions and role-to-permission mappings live in code (`backend/app/core/auth/permissions.py`) with a static `ROLE_PERMISSIONS` dict. This prevents clinics from customizing roles/permissions and forces a deploy for any change.

## Goal

Move permissions and roles into the database so that:

- Permissions are persisted as data, not code constants.
- Clinics can create **new custom roles** and assign permissions to them.
- Each module **declares its own features and default permissions** for the standard roles (`admin`, `dentist`, `hygienist`, `assistant`, `receptionist`).
- Clinics can **override the defaults** (grant/revoke specific permissions per role) without affecting other clinics.
- Modular architecture is respected: no cross-module coupling, no tech debt, no hacks.

### Data model

- `permissions` — catalog of all permissions declared by modules. Seeded/upserted on module load.
  - `id`, `module`, `code` (e.g. `clinical.patients.read`), `description`
- `roles` — roles per clinic, including system defaults.
  - `id`, `clinic_id` (nullable for system defaults), `name`, `is_system` (bool), `description`
- `role_permissions` — which permissions a role has.
  - `role_id`, `permission_id`
- `clinic_role_overrides` (optional) — per-clinic overrides of system role defaults, to avoid cloning defaults on clinic creation.
- `clinic_memberships.role_id` — replace current string `role` column with a FK to `roles`.

### Module contract

Extend `BaseModule` so each module declares:

- Its permissions (already partially supported via `get_permissions()`).
- Default permission mappings for each **standard role** (new: `get_default_role_permissions()` or similar).

On startup / migration, the registry upserts permissions and default role bindings into the DB (idempotent).

### Runtime

- `has_permission()` / `require_permission()` read from DB (with caching — per-request or short TTL in-memory) instead of the static dict.
- Wildcards (`*`, `module.*`) still supported, stored as rows or resolved at expansion time.
- `/auth/me` keeps returning the expanded permission list (frontend contract unchanged).

### Admin UI (follow-up, not in this issue)

- List roles for current clinic.
- Create/edit custom roles.
- Toggle permissions per role (showing module groupings).

## Non-goals / constraints

- **No breaking changes** to the frontend permission contract (`PERMISSIONS` config + `usePermissions`). The shape of `/me.permissions` must stay the same.
[...]

PR #400, opened 2026-09-06
feat(auth): DB-backed clinic-aware RBAC with per-clinic roles (issue #46)

lamanji:
## Overview

- Moves RBAC from code to the database with per-clinic customizable
  roles: new core tables (`roles`, `permissions`, `role_permissions`,
  `module_default_role_permissions`, `clinic_role_overrides`) via core
  migration `0007` (chained on `0006`), a DB-backed resolution switch,
  a roles management API, a seeder, and ADR 0024.
- Rebuilt fresh off `upstream/main` (the old branch predated 7 merges):
  only the 4 own commits cherry-picked. Rebase collisions found and
  fixed: ADR number (TOTP owns `0022` on main, ours is `0024`) and a
  `permissions.ts` adjacency (no duplicates).

## Tasks done

- `feat(auth)` x3: RBAC persistence foundation, DB-backed resolution
  switch, clinic-scoped role and permission management API
- `docs(auth)`: ADR 0024 + frontend `admin.roles.*` permission mirror
- `fix(auth)`: ADR renumber + reviewer blockers (seeder idempotency,
  override 422s, UUID hints, roles 403 test)

## Modules touched

- Core auth only (`backend/app/core/auth/`). No module changes, no
  catalog drift (`admin.roles.*` are core grants).

## Checklist

- [x] PR title follows Conventional Commits
- [x] `scripts/dev-verify.sh --fast` green (8 pass; only the known MSYS
  `modules-json-freshness` toolchain flake, CI-authoritative) +
  `docs-layout` green after the ADR fix + `typecheck:layers` green
- [x] Targeted container suite 30/30: `test_rbac_database_parity`,
  `test_rbac_database_switch`, `test_roles_api` (incl. new 403 test),
  `test_alembic_branches` (incl. branch heads with `0007`)
- [x] Reviewer pass: 2 BLOCKs confirmed real and fixed (seeder
  UUID-key idempotency, override None-guard 500s) + 4 shoulds fixed;
  2 nits accepted as-is (PUT update semantics kept for clients,
  singleton cache default)

## Test plan

- `db upgrade` to `0007`, open `/api/v1/roles` as admin: catalog lists
  codes, create/edit/delete a clinic-custom role, set an override on a
  system role, confirm a receptionist gets 403 everywhere.
- Flip `RBAC_FROM_DB=True`, restart, confirm `/me` permissions resolve
  from the DB identically.

## Closes

- Closes #46.

## Review

`@martinezsalmeron`: ready for review/merge when convenient.

PR #415, opened 2026-09-07
feat(auth): RBAC step 4 - role_id writes and flag-aware callers (issue #46)

lamanji:
## Overview

- Step 4 of issue #46: every membership write persists the `roles`-row FK
  (`role_id`) alongside the role string, and every permission caller
  outside `require_permission` resolves flag-aware, so custom roles and
  per-clinic overrides take effect on all backend paths under
  `RBAC_FROM_DB`.
- The flag itself and the static map stay (always-DB flip asked
  separately in the issue); flag-off behavior is unchanged.

## Tasks done

- `feat(auth)`: `resolve_role_id` + `granted_permissions_for` +
  `has_permission_for` in `core/auth/rbac.py`; `require_permission`
  and `/me` routed through them (zero behavior change)
- `feat(auth)`: `role_id` on setup register, create-user, update-user,
  and both migration-import mappers; delete-409 guard checks string
  or FK
- `feat(schedules, verifactu, copilot)`: professional gate (now async),
  prod-promote check, history/nudges/pending/digest/SSE contexts, and
  sidebar nav filtering all flag-aware
- `fix(auth)`: review round - stream-owned SSE sessions, seeder
  `role_id`, deterministic 409 test, rename-guard `or_()` mirror;
  merged current main in (no force)
- `test(auth)`: `tests/test_rbac_step4.py`, 13 tests (FK persistence,
  FK-aware delete guard, flag-on caller behavior with custom role)
- `docs(auth)`: core-api section 9b, 4 module CHANGELOG entries
  (ADR 0029 split to #438)

## Modules touched

- Core (`core/auth`, `core/plugins`); callers in schedules, verifactu,
  copilot, migration_import. No manifest, migration, permission, or
  event changes.

## Checklist

- [x] Container pytest: 13/13 new, 165 neighbors, 38 re-run green
- [x] docs-layout OK, docs-coverage strict PASS, catalog check clean
- [x] Reviewer pass: 0 functional blocks
- [x] No cross-module imports (core helpers only, isolation holds)

## Test plan

- With `RBAC_FROM_DB=false`: full existing RBAC suites green, behavior
  identical (parity + switch + roles API + permissions).
- With `RBAC_FROM_DB=true`: assign a custom role holding
  `schedules.professional.read`; professional-hours read succeeds,
  write still 403s; sidebar shows only granted entries; copilot nudges
  filter to the effective grant set.
- Delete a custom role while a member holds it by FK: 409 with the
  reassign-first message.

## Refs

- Part of #46 (step 4 of the maintainer review; scope question for the
  always-DB flip asked in the issue).

## Review


ZoliQua (changes requested):
Read the whole diff; the RBAC logic itself is sound and every new query stays inside the clinic, so this is close. A batched list:

- `backend/app/modules/copilot/router.py` (`send_message`, `confirm_tool`): please drop the `db: Depends(get_db)` parameter and compute `permissions = await granted_permissions_for(db, clinic_id, role)` as the first line inside `factory(db)`. The module docstring deliberately keeps request-scoped sessions off the streaming endpoints; with FastAPI's default request scope for generator dependencies the `get_db` session now stays checked out for the whole SSE stream. Then the matching "accepted trade-off" bullet can go from ADR 0029.
- `backend/scripts/seed_demo.py:222`: the demo seeder still writes memberships without `role_id`, so the "every membership write persists the FK" claim does not hold for demo data. `role_id=await resolve_role_id(db, CLINIC_ID, user_data["role"])` (with `seed_rbac` run first) closes it.
- `backend/tests/test_rbac_step4.py` (the 409 test): this adds a second membership for the signed-in admin in the same clinic; `get_clinic_context` takes `memberships[0]` with no ordering, so the request can resolve as receptionist and answer 403. Create the FK holder through `POST /api/v1/auth/users` (or a second user) and set that row's `role_id`.
- `router_roles.py` rename: mirror the delete guard with `or_(ClinicMembership.role == role.name, ClinicMembership.role_id == role.id)` so an FK-held membership with a drifted string is not left stale.
- `dependencies.py`: move `from .rbac import has_permission_for` back to the module top (the previous import lived there without a cycle).
- `router.py` `/me`: `granted_permissions_for` is already flag-aware; the `if settings.RBAC_FROM_DB` branch can collapse to one call.
- ADR 0029: the always-DB flip changes behaviour for every deployment, so please land it as its own PR rather than a later commit here; this one stays flag-neutral as described.
- Root `CHANGELOG.md`: one `[Unreleased]` line for step 4, as #400 got for the foundation.

Ramón's review is required on this one anyway (core auth and ADR paths).

ZoliQua (changes requested):
Thanks for the round — points 1, 2, 3, 5, 6, 7 (the ADR now lives in #438) and 8 verified on `5cb47491`. One is still open, and one thing about the diff:

- **Point 4 was about the rename**, not the delete. `update_role` (`router_roles.py` ~301-305) still matches on `ClinicMembership.role == role.name` only; the delete guard was already fine and is now `or_()`. Please mirror it: `or_(ClinicMembership.role == role.name, ClinicMembership.role_id == role.id)`, so an FK-held membership with a drifted string is not left stale on rename.
- The fold-merge (`5cb47491` = `2e143013` + the old fork head) makes GitHub show a 172-file / ~11.8k-line diff (sdi_it, sms_gateway, reports…), while the real change against main is 21 files, +547/−50. A rebase onto main would make Ramón's review much easier. The body still lists "ADR 0029 (proposed)" under tasks done — stale after the split.

`ruff check`/`format --check` clean on the changed Python; the new test selects on `ClinicMembership` carry `clinic_id`. After the rename fix this is fine from my side; `core/auth` and `core/plugins` need @martinezsalmeron's approval either way. (Small thing: the previous review was mine — tag @ZoliQua for the next round.)

ZoliQua (dismissed):
Verified on `ce6eb82e`: `update_role` now mirrors the delete guard — `or_(ClinicMembership.role == role.name, ClinicMembership.role_id == role.id)` under the `clinic_id` filter (`router_roles.py:302-312`), with a comment saying why. The upstream merge also moved the merge base, so the GitHub diff is now the real 21 files / +556/−52 instead of the old 172-file fold, and the body no longer carries the ADR line (it is in #438, merged). `ruff check` and `ruff format --check` are clean on the changed Python, and the only new production statement is that `UPDATE`, which stays inside the clinic.

One optional ask: the rename path still has no test, while the delete guard has `test_delete_role_blocked_by_role_id_holders`. A sibling case — membership held by `role_id` with a drifted `role` string, rename, assert the string follows — would keep the bug from coming back. Happy either way.

Approving from my side; `backend/app/core/**` is CODEOWNERS, so @martinezsalmeron's approval is still needed to merge. (Small thing: tag @ZoliQua rather than the lead when a round answers my review — I nearly missed this one.)

martinezsalmeron (approved):
Reviewed the full diff plus ZoliQua's rounds. Flag-aware helpers are the single entry point, every new query stays clinic-scoped, all membership writes now carry role_id, no cross-module imports. Ran the RBAC suites (73 green) against a test DB and smoke-tested users/schedules/copilot in the browser with the flag off. Merged current main in (CHANGELOG conflict only).

Issue #422, opened 2026-09-08
billing: translate the invoice PDF labels into fr/pt/de/hu/pl/it/ar


ZoliQua:
The invoice PDF (`backend/app/modules/billing/pdf.py::_get_labels`) has label sets for `es`, `en` and `ta` only. Since the fix that accepts every UI locale on `GET /billing/invoices/{id}/pdf` (the old `^(es|en|ta)$` pattern answered 422 to a pt/pl/it/de/ar UI), the other seven languages get English labels with locale-formatted amounts and dates.

**To do:** add `labels_fr`, `labels_pt`, `labels_de`, `labels_hu`, `labels_pl`, `labels_it`, `labels_ar` (about 45 keys each, including the `status` sub-map), select them in `_get_labels`, and extend `backend/tests/modules/billing/test_invoice_pdf_format.py` so each locale renders its own "Invoice" heading. The billing Nuxt layer's `i18n/locales/<code>.json` already carries most of these strings under `invoice.*`; reuse them so the PDF and the screen agree. Arabic also needs `dir="rtl"` on the `<html>` element and right-aligned amount columns.

PR #423, opened 2026-09-08
fix(billing): accept every UI locale on the invoice PDF endpoints instead of es|en|ta (#422)

ZoliQua:
Spotted by the lead while testing #394: `GET /billing/invoices/{id}/pdf` and `/pdf/preview` validated `locale` against `^(es|en|ta)$`, and the download button sends the UI language, so a pt/pl/it/de/hu/fr/ar UI got a 422 on "Download PDF" — a silent failure in seven of the ten languages we ship.

- `pdf.py` exports `PDF_LOCALES` (the ten host locales) and `PDF_LOCALE_PATTERN`; both endpoints use it. Amounts and dates already format per locale (`_LOCALE_BY_LANG` extended with fr/pt/de/hu/pl/it/ar); labels fall back to English for the languages without a label set, which is the known ceiling — #422 tracks the seven label sets, reusing the billing layer's `invoice.*` strings.
- Test: every locale in the host `i18n.locales` block of `frontend/nuxt.config.ts` must be in `PDF_LOCALES` (so adding a UI language cannot reintroduce the 422), every locale renders, `de` gets German separators with English labels for now.
- Billing `CHANGELOG.md` under Unreleased.

Small on purpose; ~30 lines of code plus the test.


martinezsalmeron (approved):
Reviewed with `main` merged (the CHANGELOG conflict is resolved in the merge commit I pushed to the branch) and checked against the running stack: `GET /billing/invoices/{id}/pdf?locale=` → 200 `application/pdf` for `de`, `ar`, `hu`; `xx` → 422. Babel renders `ar` with Latin digits and RTL marks, `hu_HU` with `1 234,50 EUR` — fine as a first step until #422 adds the label sets.

One fix pushed (3e9b6f5c): the guard test read `<repo>/frontend/nuxt.config.ts`, which only exists on a full checkout — inside docker-compose the host frontend is mounted at `/host_frontend` (`settings.DENTALPIN_FRONTEND_ROOT`), so the test raised `FileNotFoundError` locally. It now tries the setting first, the repo path second, and skips if neither exists.

The budget module has the same `^(es|en)$` pattern with pages that pass `locale.value` — filed as #441 (same shape as this PR; budget can't import billing's tuple, so mirror it or lift it to core).

Issue #424, opened 2026-09-08
settings: hydration mismatches on category pages from client-registered module sections


ZoliQua:
Seen on `/settings/billing` (dev server, Vue hydration warnings):

- `Hydration children mismatch on div.mt-6.space-y-4` at `<ModuleSlot name="settings.sections" category-filter="billing">` — the server renders fewer children than the client, because module layers register their settings sections from `*.client.ts` plugins, so SSR has no sections and the client has all of them.
- `Hydration node mismatch … expected on client: a` at `<NuxtLink to="/settings/billing/budget-expiry">` inside `<SettingsSection key="budget-expiry">` and a `children mismatch` on the `grid-cols-1 md:grid-cols-2` grid — the same pattern for host-side sections gated by client-only state.

The rail/mobile-nav mismatch on the same pages was fixed in the PR that references this issue; these two are architectural: either register slot contributions in universal plugins (so SSR and client agree), or wrap the slot output in `<ClientOnly>` with a stable placeholder. Reproduce: `npx nuxt dev`, log in, open `/settings/billing`, look at the console.

ZoliQua:
Same root cause on the dashboard (`/`, `frontend/app/pages/index.vue`): `heroEntries` / `timelineEntries` / `attentionEntries` / `widgetEntries` come from slot registrations made in module `*.client.ts` plugins, so SSR renders the `!hasAnyContent` card and the client renders three `<section>`s — four `Hydration node mismatch` warnings on every dashboard load (dev server, Vue warnings: "rendered on server: #comment, expected on client: section" ×3 and the reverse for the empty-state card). Whatever fix lands for the settings sections should cover the dashboard slots too.

ZoliQua:
PR: #458, with the before/after measured on a dev server.

I took the second of the two options in the description — keeping the rendering client-only — rather than the first. The reason is the count: there are **38** `*.client.ts` slot plugins in the tree and not a single universal one. Registering universally means every one of those plugins *and* every component they register has to be SSR-safe, and a single component reaching for `window` or `localStorage` at setup turns a hydration warning into a 500. That is a project-wide decision, so I left it here rather than making it inside a bug fix.

[...]

PR #425, opened 2026-09-08
fix(settings): settings rail hydration mismatch — gate by CSS, not the SSR-false media query

ZoliQua:
Every settings page logged "Hydration completed but contains mismatches" (found during the browser pass for #394; pre-existing on `main`). `SettingsLayout` rendered its rail with `v-if="!hideRail && isDesktop"`, and `useBreakpoint().isDesktop` is a VueUse `useMediaQuery` ref, which is `false` during SSR: the server emitted a comment where the client expected an `<aside>`. Same pattern on the mobile category nav in `settings/[category]/index.vue` (`v-if="!isDesktop && showOnboarding"`, mismatching in the other direction on desktop).

Both now render the element unconditionally and hide it with Tailwind at the same 1024 px breakpoint (`hidden lg:block` / `lg:hidden`), so server and client markup agree; the media-query refs are no longer needed in either component. Verified on the dev server: `/settings/account` loads with no hydration warning.

What remains on `/settings/billing` is a different, architectural mismatch (module settings sections registered from client-only plugins), filed as #424.

~10 lines; `frontend/app/**` so it needs the lead's review.

### DIFF
--- a/frontend/app/components/settings/SettingsLayout.vue
+++ b/frontend/app/components/settings/SettingsLayout.vue
@@ -28,8 +28,6 @@ const props = withDefaults(defineProps<Props>(), {
 })
 
 const { t } = useI18n()
-const { isDesktop } = useBreakpoint()
-
 const headerTitle = computed(() => props.title ?? t('settings.title'))
 </script>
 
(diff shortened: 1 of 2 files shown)


martinezsalmeron (approved):
Checked on the dev server with `main` merged: `/settings/account` hydrates with no warning (console filtered on `Hydration` is empty; before this the "completed but contains mismatches" error fired on every settings page). The rail is in the SSR markup with `hidden lg:block`, so server and client agree and Tailwind's `lg` (1024 px) matches the old `useMediaQuery('(min-width: 1024px)')` breakpoint exactly. eslint clean.

Unrelated, noticed on the way: `/login` still logs an attribute mismatch (`autocomplete="off"` on the server vs `email`/`current-password` on the client — Nuxt UI's `UInput` sets it client-side). Not this PR's; worth a separate one-liner if you want it.

Issue #232, opened 2026-08-21
Navigation & settings reorganization: grouped sidebar + settings super-groups (product decision)


martinezsalmeron:
Roadmap phase 17. Proposal: group the sidebar into sections and wrap Settings' 9 fixed categories into 5 named super-groups.

This is a product/UX decision, not a module — discuss mockups here before any code. Constraints:
- Navigation is built by `useModules` from module manifests; grouping must stay data-driven, not hardcoded
- Mobile-first: the grouped sidebar must be designed for mobile viewports, not just desktop

(The "free AI tier" part of this phase is out of scope — the copilot provider setup is a separate concern.)

lamanji:
**Proposal — grouped sidebar + 5 settings super-groups, fully data-driven.**

**1. Nav model.** Add an optional `section` field to every manifest `frontend.navigation` item. Sections: `care`, `lab`, `financials`, `inventory`, `practice`; headers expandable but not clickable; items without `section` render flat (Home, Schedule). `GET /modules/-/active` already curates nav by role, so only visible children show — empty groups auto-collapse.

| § | Items |
|---|---|
| *(flat, top)* | Home · **Schedule** *(renamed from "Appointments")* |
| **Clinical** | Patients · Recalls · Treatment plans · Documents |
| **Lab** | Order form · Order status |
| **Financials** | **Quotes** *(renamed from "Budgets")* · Invoices · Payments · Accountant export · Expenses · **India GST** *(moved under Financials)* · payroll (→ here when built) |
| **Inventory & supply** | Contacts · Inventory · Purchase orders · Reorder suggestions (→ here when built) |
| **Practice management** | Reports · Tasks · Staff activity |
| *(none — Copilot)* | Sidebar entry removed; already reachable via the floating AI button (bottom-right, `app.overlays`) |

**2. Settings — 9 categories → 5 super-groups** (presentation-only; URLs/routes unchanged):

| Super-group | Categories |
|---|---|
| **Clinical setup** | General · Workspace · People |
| **Clinical management** | Clinical |
| **Financial configuration** | Billing & taxes |
| **System & add-ons** | Communication · Integrations · Modules |
| **My preferences** | Account |

[...]

PR #429, opened 2026-09-09
feat(nav): optional section grouping for sidebar entries (issue #232)

lamanji:
## Overview

- Implements the approved-in-principle half of #232: optional `section`
  on manifest `frontend.navigation` items, rendered as collapsible,
  non-clickable headers by the host. Items without a section stay flat;
  empty groups collapse on their own. Nothing is hardcoded in the host.
- Mobile-first regrouping stays parked for the maintainer's mockup
  verdict; no module declares a section yet, so the sidebar renders
  exactly as before until one does.

## Tasks done

- `feat(nav)`: `section?: string` on `NavigationItem`, pure
  `groupNavigationItems` helper, grouped sidebar + drawer rendering
  (collapsed icon sidebar stays flat, settings stays separated)
- `test(nav)`: `tests/utils/navigation.test.ts` (flat order, group
  order, empty list)
- `docs(modules)`: `section` row in the creating-modules navigation
  example

## Modules touched

- Host frontend + module-author guide only. No backend, manifest,
  permission, locale, or generated-file changes (backend passes the
  `frontend` dict through untouched).

## Checklist

- [x] `vitest` full unit suite 142/142 (3 new), `eslint` clean
- [x] `typecheck:layers` blocked by a pre-existing vue-router/volar
  toolchain break (fails at plugin init, none of our files named;
  same class as the documented local-only flakes - CI arbitrates)
- [x] Reviewer pass: 0 blockers
- [x] M5-clean: single commit on current `upstream/main`

## Test plan

- Sidebar and drawer render byte-identical to main (no sectioned
  items exist yet). Declare `"section": "clinical"` on any module nav
  item locally: a collapsible header appears in sidebar + drawer,
  collapses/expands with `aria-expanded`, vanishes for roles that
  cannot see any of its items.

## Refs

- Part of #232 (approved mechanism only; mobile regrouping parked).

## Review


ZoliQua (commented):
The mechanism reads well and it is genuinely data-driven: `groupNavigationItems` is pure and tested, `section` is optional on the manifest contract and, with no module declaring it yet, the sidebar renders exactly as before. Before this can go to Ramón (`frontend/app/**`), a few points:

- #232's last maintainer comment asked for a "go" on the mobile mockups before code; the mockups are posted but the go is not, and this PR does implement the mobile drawer grouping. Could you get that on the issue first?
- `frontend/app/layouts/default.vue` — the drawer section header is `px-3 pt-3 pb-1 text-caption` (~32 px); the mockups promised 44 px rows like the items (`py-3` + 20 px icon). `py-3` or `min-h-11` on the header.
- The layout now carries four copies of the `NuxtLink` block (sidebar/drawer × flat/grouped). Please fold them into one small component or a single loop over `[...flat, ...groups]` with a `section?` discriminator.
- Collapsed state is `useState('sidebar:nav:collapsed')` (per tab, lost on reload) while `isSidebarCollapsed` persists to `localStorage('sidebar:collapsed')`. Persist the same way, or say why not.
- `nav.sections.<key>` has no owner: no locale defines any `nav.sections.*`, and the keys are shared by many modules, so they must live in the host locales. Shipping the five approved keys (`clinical`, `lab`, `financials`, `inventory`, `practice`) in the ten host locales here avoids a separate parity PR for the first module that uses it.
- a11y: `aria-expanded` is there; add `aria-controls`/`id` on header/panel, and check the `uppercase text-subtle` header contrast in dark mode.
- Nit: grouping lives in `utils/navigation.ts`, not `useModules` as the issue wording had it — worth one line in the description.

ZoliQua (changes requested):
Verified on `8f06bcfa`: drawer and sidebar headers at `py-3` (2), the four copies folded into `components/nav/NavRow.vue` and used at all four call sites (3), `nav.sections.*` present with identical key sets in all ten host locales — scripted diff, 2507 keys each, 0 missing (5), `aria-controls`/`id` pairs plus `role="region"` (6), description (7). `tests/utils/navigation.test.ts` still covers flat order, group order and the empty list.

Two things before I can approve:

1. **`sidebar:nav:collapsed` is written but never read back.** The read sits inside the `useState` initializer guarded by `import.meta.client`; under SSR the initializer runs on the server, returns `{}`, and the client hydrates that payload without re-running it, so a reload forgets the collapsed sections. `isSidebarCollapsed` avoids this by loading in `onMounted` — please do the same here.
2. `frontend/i18n/locales/en.json` lost its trailing newline in this commit (the other nine keep it).

And the #232 mobile "go" is still pending from Ramón (his last comment asks for the mockups; yours are up, no reply yet). Since this PR ships the drawer grouping, it waits on that plus `frontend/app/**` CODEOWNERS. The previous review was mine, so tag @ZoliQua for the next round.

ZoliQua (dismissed):
Both points verified on `cd6a96e8`: the collapsed map now loads in `onMounted` with the `useState` initializer left as `() => ({})`, so the SSR payload no longer wins on hydration (`layouts/default.vue:79-87`), and all ten locale files end with a newline again. The round-2 fixes are all intact — `NavRow` at the four call sites, `py-3` headers, `aria-controls`/`id` pairs with distinct sidebar and drawer prefixes, and identical `nav.sections.*` key sets across the ten host locales.

The code is ready from my side, so I am approving it, but please do not read that as clearance to merge: it still waits on @martinezsalmeron for the mobile drawer go in #232 (his 2026-09-04 comment is still the last word there) and for `frontend/app/**` CODEOWNERS.

Issue #63, opened 2026-04-28
WhatsApp notifications via pluggable outbound gateway (Zapier / Make / Cloud API)


martinezsalmeron:
## Problem

In Spain (and most of LATAM) patients read WhatsApp; they often ignore email and SMS. Today every outbound notification — appointment confirmation, 24 h reminder, recall, post-op follow-up — goes through email only. Confirmation rates and recall conversion suffer because the channel is wrong.

We want **all clinic-to-patient notifications to be sendable via WhatsApp when the patient has a WhatsApp-reachable phone number**, with email as fallback when not.

## Strategy: outbound gateway with pluggable adapters (don't lock into one vendor)

Direct integration with the Meta WhatsApp Cloud API is powerful but heavy: business verification, template approval, 24 h session window rules, per-message cost. A clinic that just wants to start sending messages today should not need to go through that.

Plan: build a generic **Notification Gateway** in core, with pluggable channel adapters. Ship multiple adapters so each clinic picks what fits:

| Adapter | Effort for clinic | Notes |
|---------|------------------|-------|
| `webhook` (Zapier / Make / n8n) | Lowest — paste a webhook URL | Clinic owns the Zap; we POST a JSON payload, they route it to WhatsApp / SMS / Telegram / anything. Best default. |
| `whatsapp_cloud_api` | Medium — Meta business verification + template approval | Direct, no third-party hop, lower per-message cost at scale. |
| `twilio` / `360dialog` / `messagebird` | Medium — BSP account | Easier WhatsApp onboarding than Meta direct, paid. |
| `click_to_whatsapp` (no API) | None | Generates `https://wa.me/<phone>?text=<msg>` links the clinic clicks to send manually. Useful for tiny clinics with no budget. |
| `email` | Already in core | Fallback channel. |
| `sms` (future) | Via BSP | Fallback when no WhatsApp. |

The gateway is core; each adapter beyond `webhook` and `email` ships as its own **community module**, so clinics install only what they use.

### 1. Notification Gateway (core)

New module-agnostic service responsible for delivering a notification through the patient's preferred channel with fallbacks.

[...]

PR #430, opened 2026-09-09
feat(notifications): built-in WebPush channel + VAPID (issue #63)

lamanji:
## Overview

- Built-in WebPush channel for the notifications gateway (T6): VAPID
  keypair per deployment, per-patient browser subscription registry,
  fan-out delivery with dead-endpoint pruning, opt-out preferences.
- Rebuilt M5-clean off current main from the local-only branch; the
  SMS rework that landed meanwhile was merged in (enum, gateway
  branch, preference columns coexist), and the migration rides as
  `notif_0006` chained on the SMS `notif_0005`.

## Tasks done

- `feat(notifications)`: `PushAdapter` + VAPID helpers + subscription
  service, gateway PUSH branch, subscription + public-key endpoints,
  `push_enabled` preference, `push.read`/`push.write` permissions
- `fix(notifications)`: migration renamed `notif_0006`, chained on SMS
  revision (was colliding on `notif_0005`)
- `chore(deps)`: `pywebpush` locked via `uv lock` (additive only)
- `test(notifications)`: `test_push_channel.py` (adapter, registry,
  pruning, opt-out)
- `docs(notifications)`: permissions row + module CHANGELOG/CLAUDE.md

## Modules touched

- `notifications` only (+ host `app/config.py` VAPID settings,
  `pyproject.toml`/`uv.lock`, host `permissions.ts`). No new module,
  no cross-module imports. VAPID private key is env-only, never
  committed, never logged.

## Checklist

- [x] Container pytest `test_push_channel.py` 11/11 green (rebased
  onto post-merge main, `notif_0006` chained on SMS `notif_0005`)
- [x] Revision chain `notif_0001`-`notif_0006` unique, model indexes
  match the migration
- [x] Reviewer pass: 0 blockers (+ first-party L1/L33/secret checks)
- [x] M5-clean: merge-base is current `upstream/main`

## Test plan

- Set `DENTALPIN_VAPID_PRIVATE_KEY` + subject, register a browser
  subscription via `POST /api/v1/notifications/push/subscriptions`,
  enqueue a notification: push delivered, dead endpoints pruned on
  410/404, `push_enabled=false` opts out. Without the key the channel
  resolves nothing (SMS/email fallback unaffected).

## Refs

- Refs #63 (gateway alignment; browser service worker stays a
  follow-up).

## Review


ZoliQua (changes requested):
Thanks — clean module wiring (registry via `on_activate`, own migration chain `notif_0006` on `notif_0005`, service/router split, clinic filter on every query incl. pruning, `uv.lock` regenerated with pywebpush 2.5.0 / py-vapid / http-ece, licences fine). Before this can go in I need the points below; the first one I am escalating to Ramón because it is a product/architecture call.

1. **Patient subscribe path (design, needs the lead).** All four routes require a staff JWT with `notifications.push.write`, and browser subscriptions are per browser, so only staff browsers can ever register — yet rows are keyed by `patient_id` and routed through patient consent. Without a patient-facing subscribe flow the channel cannot reach a patient end to end, and we do not merge scaffolding for a follow-up. Either bring the service worker + a patient-token subscribe endpoint into this PR (that surface needs Ramón's OK), or describe the intended end-to-end flow so we can decide.
2. `channels/push_adapter.py:82` — `pywebpush.webpush()` is synchronous (`requests`) and defaults `timeout` to 10 000 s; called from the async dispatcher, one stalled push service freezes the whole app. Please use `webpush_async` (aiohttp is already locked) or `asyncio.to_thread`, and pass `timeout=10`.
3. TTL and payload size: pass `ttl` (24 h is sensible for reminders; the default 0 drops the message if the device is offline) and cap the JSON payload under 4 KB before sending.
4. `schemas.py:382` / `push.py:35` — the endpoint is only checked for non-emptiness; the worker POSTs wherever a staff user pointed it (`http://10.0.0.5/…`). Require `https://` in the schema and the service.
5. VAPID config: `vapid.py` reads `os.environ` while `config.py` adds unused `Settings` fields; a key in `.env` outside Docker is silently ignored. Read `settings` (and cache the parsed key) or drop the `config.py` change.
6. `gateway.py` — the push row's `to_address` is the patient's e-mail; the logs view shows it as the recipient. Something like `push:<n>` reads right.
7. `push_adapter.py` — import `webpush` at module top (hard dependency) and patch it in tests instead of a `sys.modules` stub; use `WebPushException.status_code` in `_is_gone`; make the test stub `**kwargs`-tolerant; avoid `unregister("webpush")` in fixture teardown (it removes the adapter `on_activate` registered).
8. Model/migration drift: `patient_id index=True` and the `clinic_id` index name do not match `notif_0006`.
9. Docs/scope: `docs/technical/notifications/overview.md` (channels, API surface, permissions) is not updated; `pushRead/pushWrite` in `frontend/app/config/permissions.ts` are referenced nowhere — drop them until there is a UI, which also removes the CODEOWNERS path.
10. Tests: add a cross-clinic case (list/delete from clinic B → 404; pruning never touches B).

ZoliQua (changes requested):
Thanks for the fast round — 2, 3, 6, 8 and 9 are done on `591d527a`, the model/migration indexes match, `notif_0007` downgrades cleanly, locale parity holds in all ten layer files, and the token flow is a clean shape. Three things still block, one of them a real bug:

1. **Send is broken in production.** `push_adapter.py` (~line 100) passes `vapid_private_key=vapid_config.vapid_private_key()` — the PEM string. pywebpush routes a string through `py_vapid.Vapid.from_string`, which only accepts base64url RAW/DER; with 2.5.0 that raises `ValueError: Could not deserialize key data`. It lands in the generic `except Exception`, gets logged as "webpush send failed" and the row is `FAILED` — silently, for every subscription. The tests stub `webpush_async`, so they never see it. Please pass a cached `Vapid.from_pem(...)` instance (you already cache the parsed key) and add one test that runs the real `webpush_async` with the HTTP call patched, so the VAPID path is covered. `supports()` should also depend on `vapid_public_key() is not None` rather than "non-empty string".
2. **Point 4:** the `https://` check is only in the service; `PushSubscriptionCreate.endpoint` / `PushPatientRedeem.endpoint` need the validator too, with a test on the public route.
3. **Point 7:** please make the `webpush` import hard. CI's manifest job installs `.[dev]` (`ci.yml` ~110), so the "host tooling without the dep" fallback is not needed, and it adds a runtime "not installed" branch.
4. **Point 10:** the cross-clinic test is service-level only; add the HTTP list/delete-from-clinic-B → 404 case and a pruning case (a 410 on A never deletes B's row).
5. **Public router hardening:** the budget public router uses `limiter` (`60/minute` GET, per-token `5/15minute` POST); `notifications/public_router.py` has none. Return `{subscribed: true}` on redeem instead of `id`/`clinic_id`/`patient_id`; make the single-use redeem atomic (`UPDATE … WHERE used_at IS NULL RETURNING`, today it is read-check-write).
6. Small: the 4 KB cap is pre-encryption (aes128gcm adds ~100 bytes — cap around 3990, and `ensure_ascii=False`); `await navigator.serviceWorker.ready` before `pushManager.subscribe`; `notificationclick` opens `/` which is the staff login for a patient; the screen doc says staff mints the token "from the patient file" — there is no UI, say it is API-only for now.

`frontend/app/middleware/auth.global.ts` is CODEOWNERS, and the public route + root-scoped service worker is the surface I asked Ramón about — holding for @martinezsalmeron after the fixes. The previous review was mine; tag @ZoliQua for the next round.

ZoliQua (changes requested):
This round moved most of it, and I checked the VAPID fix end to end rather than from the diff: `vapid.py` builds `Vapid.from_pem(...)` behind an `lru_cache` keyed by the PEM, `push_adapter.py` passes the instance, and with the locked pywebpush 2.5.0 that takes the `isinstance(vapid_private_key, Vapid01)` branch, so it signs. `supports()` now keys off `vapid_public_key()`. Point 2 (validator on both schemas plus the public-route test), point 5 (limiter matching the budget public router, `{subscribed: true}`, and a genuinely atomic `UPDATE … WHERE used_at IS NULL RETURNING`) and the pruning half of point 10 are done.

Four things left, two of them real:

1. **Point 3 is not in the commit.** `push_adapter.py:20-26` still has the `try/except ImportError` fallback and the "pywebpush is not installed" runtime branch is still there, although `9482e768`'s message lists the hard import. CI installs `.[dev]`, so please make it a plain top-level import and drop the branch.
2. **New bug: `_fit_payload` re-serialises without `ensure_ascii=False`.** Truncation is the only place the cap matters, and it is exactly where the escaping comes back, so the function can return a payload larger than the one it was given. Ran it: a 4000-character accented body goes in at 8026 bytes and comes out at **11909** against the 3990 cap, which the push service answers with 413. One keyword argument fixes it; a test with a long accented body would pin it.
3. **Point 10's HTTP half.** `test_cross_clinic_push_http_404s` ends by deleting clinic A's own row with clinic A's headers and asserting 204, under a comment about B-scoped deletes being refused. As written it passes even with no clinic filter at all. Please create a subscription in clinic B and assert that `GET` and `DELETE` with A's headers answer 404.
4. **Point 1's test.** The `isinstance(..., Vapid)` assertion is a good guard, but nothing exercises `Vapid.sign` or pywebpush's key branch; a test calling the real `webpush_async` with only the HTTP call patched would cover the path that was broken. And the screen doc still reads "Staff mints the token from the patient file" — please say token minting is API-only for now.

Separately, `frontend/public/push-sw.js` sits in the host public directory rather than the module's frontend layer, so uninstalling notifications leaves the worker behind; worth moving if the layer's `public/` is served.

`frontend/app/middleware/auth.global.ts` is CODEOWNERS, and the public route plus the root-scoped service worker is the surface I raised with @martinezsalmeron, so this waits on him after the fixes.

ZoliQua (dismissed):
All four are in, and I checked the two that mattered by running them rather than reading them.

- **Point 1, the hard import**: `from pywebpush import WebPushException, webpush_async` at the top, no fallback branch. Worth knowing that this is now load-bearing: my review venv did not have `pywebpush` and the test module failed at collection, which is the honest behaviour for a declared dependency and exactly what the old `try/except` was hiding.
- **Point 2, the payload cap**: `_fit_payload` re-serialises with `ensure_ascii=False`. Ran the same input that broke before — a 4000-character accented body enters at 8026 bytes and now leaves at **3989** against the 3990 cap, where before it left at 11909.
- **Point 3, the cross-clinic HTTP test**: `test_cross_clinic_push_http_404s` now creates a subscription in clinic B and asserts that A's `GET` lists nothing and A's `DELETE` on B's row answers 404, with the A-scoped delete still returning 204. That is the test I asked for, and the comment explains what would happen without the filter.
- **Point 4**: the new test exercises the `isinstance(vapid_private_key, Vapid01)` branch and `Vapid.sign` end to end, and the screen doc now says token minting is API-only.

`ruff` clean, `pytest tests/modules/notifications` 30 passed locally, CI green.

Approving from my side. Two things still stand between this and merge, both for @martinezsalmeron: `frontend/app/middleware/auth.global.ts` is a CODEOWNERS path, and the public patient route plus the root-scoped service worker is the surface I asked him to look at. The service worker living in `frontend/public/` rather than the module's layer is the one design point I would still like an answer on, since uninstalling notifications leaves it behind.

PR #431, opened 2026-09-09
feat(patients, expenses): CSV import dry-run + patient restore

lamanji:
## Overview

- CSV import with dry-run validation for patients and expenses, plus a
  restore endpoint for soft-archived patients. Rebuilt M5-clean off
  current main from the local-only branch; the unrelated Sentry
  error-tracking commit was deliberately left out (separate chore,
  own decision).
- Review fixes already applied on the branch: Spanish accents
  restored in patient tool descriptions, `import_expenses` takes
  `UUID | None` like the canonical create path.

## Tasks done

- `feat(patients)`: CSV import with dry-run validation (T7a)
- `feat(expenses)`: CSV import with dry-run validation (T7b)
- `feat(patients)`: restore soft-archived patient endpoint
- `test(patients, expenses)`: import + restore suites
- `docs(patients, expenses)`: permissions rows + module
  CHANGELOG/CLAUDE entries

## Modules touched

- `patients`, `expenses` only (+ their module test dirs). Agent tools
  wrap the services with `clinic_id` filtering; no new migrations, no
  cross-module imports. No tracking issue exists for this work -
  triage welcome (no `Closes` claimed).

## Checklist

- [x] Container pytest patients + expenses 11/11 green (rebased
  onto post-merge main) + tools battery 42/42 (new restore tool)
  (1 deselected = pre-existing roundtrip exclusion)
- [x] Reviewer pass: 1 real blocker found and fixed (accents +
  type hint), re-verified first-party
- [x] M5-clean: merge-base is current `upstream/main`

## Test plan

- Import a CSV with `dry_run=true`: row-level 422 report, nothing
  persisted; `dry_run=false` commits. Archive a patient, restore via
  the endpoint: status flips to active, history intact.

## Review


ZoliQua (changes requested):
Clean structure — validation through `PatientCreate`, dry-run by default, `ctx.clinic_id` on every write and on restore, no migration needed. A few things before merge:

- **Restore vs. the media cascade.** `patient.archived` makes `media` soft-archive the patient's documents (`media/__init__.py:67`); `restore_patient` publishes only `patient.updated`, so the documents stay archived and the "history intact" claim does not hold for files. Please add `PATIENT_RESTORED` (types, `events.md`, catalog regen) with a media handler that reverses the cascade, or state the limitation explicitly and open an issue.
- **Duplicates.** Nothing stops importing the same file twice; `patients` has no uniqueness on national id / e-mail / phone, so a re-upload doubles the clinic and fires `patient.created` for each. Could the dry-run report flag rows matching an existing active patient (national id, or e-mail + date of birth) and commit skip them unless explicitly allowed? That is the difference between a demo and a migration tool (`migration_import` is idempotent through its entity mappings).
- **Delimiter and dates.** `csv.DictReader` with the comma only; Spanish Excel writes `;`-separated files with `DD/MM/YYYY`, both rejected today. `csv.Sniffer` over `,;` with a comma fallback, and a documented date format, cover most real files.
- **Commit semantics.** `ExpenseService.create_expense` commits per row (`expenses/service.py:33`) while `create_patient` only flushes, so an expenses import that fails mid-way is partially persisted and returns a 500 with no report, while patients is all-or-nothing. Align (build rows, commit once) or document both.
- **Agent tools.** The copilot cannot upload a file, so `import_*_csv` would have the LLM emit up to 1,000 rows of PII as an argument. I would drop both tools from this PR and keep the HTTP endpoints (which satisfy "API first"). If you keep them: descriptions are Spanish in patients and English in expenses, and the tool returns `created: list[UUID]` while HTTP returns `created: int`.
- `await file.read()` buffers the whole upload before the 1 MiB check — read with a limit.
- Tests: `test_patient_csv_import.py` says "isolation" in its docstring but has no second-clinic case; add restore of another clinic's patient → 404.
- Docs: `docs/user-manual/{en,es}/patients/index.md` and `expenses/index.md` need a short "Importing from CSV" section (curl with `-F file=@…`, `?dry_run=false`, columns, encoding, delimiter, date format, caps, report shape); today the column contract exists only in a module docstring.

No CODEOWNERS path, so I can merge once these are in.

ZoliQua (changes requested):
Every point from the first round is in and verified on `fe4a14c2`: `PATIENT_RESTORED` in `EventType`, documented, catalogs regenerated (`--check` green), the media handler reverses the cascade with `clinic_id` filtered and the publish is transactional (`db=db`); duplicate report + skip unless `allow_duplicates`; Sniffer over `,;` with comma fallback and both date formats documented; bulk create flushes only; the agent tools are gone; second-clinic restore → 404 tested; manuals en+es. `ruff` clean, every new `select(` carries `clinic_id`.

Two things the round introduced, then it is ready for Ramón (this PR now touches `core/events/types.py`, so CODEOWNERS requires his review):

1. **Oversize upload returns 500, not 422.** `read_upload_limited` is called *before* the `try` in both routers (`patients/router.py` ~127, `expenses/router.py` ~85), so its `CsvImportError` falls through to the global handler. Move the call inside the `try` and add a test with `MAX_CSV_BYTES + 1` bytes — there is none, which is why it passed.
2. **`find_duplicate_lines` uses `scalar_one_or_none()`** (`patients/csv_import.py` ~175, ~187). After one `allow_duplicates=true` import — or in any clinic that already has two patients with the same national id — the next dry-run raises `MultipleResultsFound`. Use `.scalars().first()`.

Optional: a line in the media docstring that the restore cascade re-activates every archived document of the patient, not only the ones the archive cascade touched; intra-file duplicates (two rows with the same id) are not flagged against each other.

The previous review was mine, so tag @ZoliQua for the next round.

ZoliQua (dismissed):
Both blockers are fixed and verified on `2341ae70`. `read_upload_limited` now runs inside the `try` in `patients/router.py:127` and `expenses/router.py:85`, and `test_oversize_upload_422s_not_500s` exercises it with a body over `MAX_CSV_BYTES`. `find_duplicate_lines` uses `.scalars().first()` on both match branches — no `scalar_one_or_none` left in the file. The intra-file duplicate pass is a nice addition. `ruff check` and `ruff format --check` clean, `generate_catalogs.py --check` green, and the eight earlier points all still hold.

Three small things, none blocking:

- `expenses/csv_import.py` has its own copy of `read_upload_limited`, so the expenses 422 path is still untested — please mirror the patients test into `test_expenses_csv.py`.
- `CsvRowDuplicate.patient_id` is now `UUID | None` for intra-file repeats; the en/es manuals still describe `duplicates` as matches against existing patients. One line about `matched_on: "same_file"` with a null `patient_id` keeps the documented contract true.
- The optional docstring note from last round: `media/service.py` says "the documents archived with the patient", but the query restores every archived document of that patient.

Approving from my side; the PR touches `backend/app/core/events/types.py`, so CODEOWNERS needs @martinezsalmeron before it can go in.

martinezsalmeron (approved):
Reviewed in depth on top of ZoliQua's three rounds. Ran the 19 new tests in the container, then a live HTTP smoke on the dev stack: `;`-delimited CSV with accents and DD/MM/YYYY, dry-run → commit → re-import (DB duplicate flagged by national_id, intra-file `same_file` flagged and skipped), oversize upload → 422, archive → restore → no-op restore → 404 on unknown id, expenses dry-run/commit.

One thing the smoke caught, fixed on the branch in 6cc3ba09: Spanish Excel writes `1200,50` next to the `;` delimiter this PR already sniffs, and `amount` rejected it. The decimal comma is now accepted (test + en/es manual + CHANGELOG).

Known, acceptable: `patient.restored` reverses only the media cascade; recalls parked in `needs_review` by the archive stay there for staff to review, which is the right default. Merging.

PR #432, opened 2026-09-09
feat(ops): db backup CLI + disaster-recovery runbook

lamanji:
## Overview

- Disaster recovery for self-hosters: a `dentalpin db backup` command
  (full `pg_dump` custom-format + storage tarball, retention pruning)
  plus a canonical runbook (nightly schedule, offsite + encryption,
  RPO/RTO, fresh-host restore, outage checklist, monthly
  verification).
- Closes the documentation gap where full-database backups were a
  one-line hand-wave while patient files live outside Postgres.

## Tasks done

- `feat(ops)`: `dentalpin db backup` (`--out-dir`, `--keep`,
  `--skip-storage`; exit 3 on missing/failed/timed-out `pg_dump`)
- `test(ops)`: `tests/test_cli_db_backup.py` (8 tests: prune, missing
  binary, failure, timeout, tarball exclusion, handler codes, parser)
- `docs(ops)`: `docs/workflows/backup-restore.md` + full-backup
  pointers in `operations.md` en/es (replaces the hand-wave line)

## Modules touched

- Core CLI only. No migrations, no RBAC surface (operator-side), no
  secrets in backups by design (`.env`/keys excluded, documented).

## Checklist

- [x] Container pytest 8/8 green
- [x] docs-layout OK
- [x] Reviewer pass: 1 valid robustness finding fixed (timeout
  handling) + test added
- [x] M5-clean: single commit on current `upstream/main`

## Test plan

- `dentalpin db backup` in the backend container: dump + tarball
  land in `storage/backups/`, older generations pruned, exit 0;
  without `pg_dump` on PATH: clean error, exit 3. Restore drill per
  the runbook §5 on a spare host.

## Review


ZoliQua (changes requested):
This fills a real gap and the runbook reads well (en/es parity, correct dump format for `pg_restore`). Blockers:

- The documented command does not exist. `docs/workflows/backup-restore.md` §1, the cron line and both `operations.md` hunks say `docker compose exec backend dentalpin db backup`, but `backend/pyproject.toml` has no `[project.scripts]`, the image installs no such binary, and the entrypoint and the rest of `operations.md` use `python -m app.cli` / `./bin/dentalpin`. Either switch every invocation to `./bin/dentalpin db backup` / `docker compose exec -T backend python -m app.cli db backup`, or add the script entry and re-lock `uv.lock`.
- `app/cli/db.py:126` — `prune_backups` uses `candidates[: max(0, len(candidates) - keep)]` with no floor on `--keep`: `--keep 0` (or a negative value) deletes the backup that was just written. Please enforce `keep >= 1`, never prune files created in this run, and add a test for both.
- `capture_output=True` holds the whole dump in memory before `write_bytes`; a multi-GB database OOMs the container. Stream with `--file=<target>` or `stdout=fh`, as the module-uninstall dump in `processor.py` already does.
- Restore recipe: `pg_restore -c` on an empty database errors on every "does not exist" and exits 1 — exactly what the runbook tells operators to alarm on. `pg_restore --clean --if-exists --no-owner -d dental_clinic` is the safe form.
- Nice-to-have: `snapshot_storage` skips the literal folder name `backups` at the top level only; skip `out_dir` by path when it is inside the storage root, or `--out-dir /app/storage/nightly` tars the tarball into itself. And `PGPASSWORD` in `env=` keeps the password off `ps`.

No CODEOWNERS path (`backend/app/cli/` is outside `core/`), so I can merge after these; the `_pg_dump_dsn` import from `core/plugins/processor.py` is a private helper that should become public in a lead-owned follow-up.

ZoliQua (changes requested):
All five points are in and match the code on `49454bd0`: every documented invocation is `docker compose exec -T backend python -m app.cli db backup` (no `dentalpin db backup` left in docs), `--keep` floored at 1 with the fresh files protected and tested, the dump streams to the file, the restore recipe has `--clean --if-exists --no-owner`, out-dir skipped by path and `PGPASSWORD` in the env. `ruff` clean.

One thing to close before I merge: on a failed or timed-out `pg_dump`, `dump_database` opens `target` before the run and returns the error without removing it, so a partial `full_<ts>.dump` stays on disk. The next run's prune treats it as the newest backup, and the runbook tells the operator to restore "the newest". Please `target.unlink(missing_ok=True)` on the failure, timeout and empty-output branches and assert absence in `test_dump_database_failure`/`_timeout`.

Two optional nits: a test that `_cmd_backup` returns 2 on `--keep 0`, and resolving paths in `snapshot_storage` so a nested `--out-dir /app/storage/x/y` is skipped too (today only a direct child is). Also worth an `unquote()` on the password pulled from the DSN — `urlparse` gives it percent-encoded, libpq decodes.

I'll do a real backup + restore on a scratch stack before merging. No CODEOWNERS path here.

ZoliQua (changes requested):
The three required changes are in and verified on `145b75e7`: `target.unlink(missing_ok=True)` on the timeout, failure and empty-output branches with `assert not target.exists()` in both tests, plus all three optional items — the `--keep 0` exit code test, the path-resolving nested out-dir skip, and `unquote()` on the DSN password. `ruff` clean, `pytest tests/test_cli_db_backup.py` 12/12 green. I also checked that `postgresql-client` is in the backend image and matches `postgres:15-alpine` in the compose files, so `pg_dump` really is on PATH and version-compatible.

Two things in the runbook block the merge, because they are in the one procedure that runs when everything else has already failed:

1. **§3 cannot be executed as written.** Step 1 starts only the database (`docker compose up -d db`), then step 3 runs `docker compose exec -T backend tar -xz …`. `exec` needs a running container, so the operator hits "service backend is not running" in the middle of a restore. `docker compose run --rm -T backend tar -xz -C /app/storage < storage_….tar.gz` works with the stack down.
2. **§4 step 3 describes success as the failure output.** It says `db upgrade` "must print no targets (a no-op)", but `_cmd_upgrade` prints `db upgrade: no Alembic targets` **and returns 1** in that case, while a healthy stack always prints a non-empty target list and applies it. As written the step misreads success as failure and silently applies migrations during a restore. Please make it `db upgrade --dry-run` and describe success as "prints the target list, exits 0".

One more I would like folded in: `snapshot_storage` has no error handling, so a permission error, a file removed mid-run or a full disk leaves a truncated `storage_<ts>.tar.gz` behind that the next prune keeps as the newest — the same bite you just fixed for the dump. Same shape: `try/except OSError`, unlink, return an error string, non-zero exit.

Nits: `--out-dir /app/storage` (the storage root itself) skips nothing, because the check only asks whether `out_dir` is inside a child — test the other direction too; `--skip-storage` also skips pruning `storage_*`, so a cron switched to it keeps every old tarball forever; and `_dsn_and_env` has no test, which is a shame for the one helper that handles the password.

Post these and I will merge — no CODEOWNERS path here, and I will run a real backup and restore on a scratch stack first.

ZoliQua (approved):
All three are in, and I ran the thing end to end on a real cluster rather than reading the diff.

Verified on `145b75e7`+: the runbook's §3 now uses `docker compose run --rm -T backend tar …`, which works with the stack down; §4 step 3 is `db upgrade --dry-run` described as printing the target list; and `snapshot_storage` returns an error instead of `None`, unlinks its partial tarball, skips the target file itself (so `--out-dir` pointing at the storage root is safe now too) and is wired to exit 4. Pruning runs for both prefixes outside the `--skip-storage` branch, and `_dsn_and_env` has its encoded-password test. `ruff` clean, 19/19 green, CI green.

The real run, against a seeded database and a storage tree with a 200 KB binary in it:

- `db backup` produced both artefacts; `pg_restore -l` lists 151 `TABLE DATA` entries and the tarball contains the storage tree.
- Restored into an empty database with the runbook's own `--clean --if-exists --no-owner` recipe: patients 15 → 15, invoices 7 → 7, public tables 151 → 151.
- `--keep 1` twice: the older pair is pruned, the pair just written survives.
- A failed dump (DSN pointing at a database that does not exist): exit 3, the error printed, and **no partial file left in the directory** — the count stayed at the previous good pair. That was the one blocking bug, and it is properly fixed.

Merging. Thanks for the three careful rounds on this one; the disaster-recovery path is the last place anyone wants a surprise.

Issue #133, opened 2026-07-30
New module: electronic invoicing for Italy (FatturaPA / SDI), the verifactu pattern applied to IT


martinezsalmeron:
## The gap

An Italian practice issues electronic invoices through the SDI (Sistema di Interscambio) every day. Dentalpin's `billing` module issues invoices with configurable series, partial payments and a branded PDF, and stops there. There is no FatturaPA XML, no transmission, no receipt handling.

That single gap is why our own [Italian comparison against XDENT](https://www.dentalpin.com/it/blog/dentalpin-vs-xdent/) tells the reader to choose the other product if they need the software to invoice. We would rather fix the gap than keep writing that sentence.

## The pattern already exists

`verifactu` is exactly this problem solved for Spain, and it is the shape to copy rather than a thing to extend:

```
backend/app/modules/verifactu/
  hook.py            # plugs into billing via BillingHookRegistry (backend/app/modules/billing/hooks.py)
  models.py          # verifactu_* tables only, no columns added to billing
  services/          # XML build, signing, transport, response parsing
  tasks.py           # periodic worker, drains pending records with back-off
  migrations/        # its own Alembic branch
  frontend/          # Nuxt layer: settings screen, queue screen, its own i18n
```

Two rules that made it work and should hold here too:

- **The module is gated by country.** `verifactu` only registers its billing hook when the clinic's country is `ES`. An `sdi_it` module registers when the country is `IT`, and a practice anywhere else never sees a screen.
- **No columns in `billing` or in the catalog.** Everything lives in `sdi_it_*` tables. Uninstalling leaves no trace, and there is a round-trip test that proves it.

`docs/modules/verifactu.md` documents the whole thing end to end.

## What to establish before writing code

This issue deliberately does not state the rules, because we have not verified them and a half-remembered obligation is worse than none. The first useful contribution is a short spec, as a comment or an ADR under `docs/adr/`, sourced from [Agenzia delle Entrate](https://www.agenziaentrate.gov.it/) and nowhere else:

1. **The invoice format and its schema version.** FatturaPA XML: which version is current, which fields a dental practice actually fills, how the *natura* codes apply to exempt healthcare services (the parallel to the `E1` mapping `verifactu` does for Spanish VAT).
2. **Transmission.** Whether a practice transmits to the SDI directly with a certificate or through an intermediary, and what that means for a self-hosted installation with no vendor in the middle. This is the decision that shapes the module.
[...]

PR #435, opened 2026-09-09
feat(sdi_it): PEC transport — SMTP send, IMAP receipt poll, worker; Imposta rounding and progressivo lock (#133)

ZoliQua:
Second `sdi_it` PR (ADR 0025 §3, after #410 and the lead's #418). Backend only; the Nuxt layer is the third PR.

**PEC transport.** With `transport = pec` the clinic's own PEC mailbox is the channel: every `pending` file is sent as an XML attachment to the SDI mailbox (`sdi01@pec.fatturapa.it` first; the SDI answers from a dedicated `sdiNN@pec.fatturapa.it` address, which the poller stores and uses from then on — spec §1.3), and the receipts come back as attachments (XML, or inside a zip). `services/pec_transport.py` is stdlib `smtplib`/`imaplib` in the default executor — no new dependency, no `uv.lock` change. `services/submission_queue.py` drains one clinic per tick: rows locked `FOR UPDATE SKIP LOCKED`, commit before every network call, per-file backoff 2 min → 1 h over eight attempts then `failed`, a mailbox-level error pauses the clinic ten minutes with the message in `last_error`; receipts are applied through `services/receipts.apply_receipt`, now shared with the API import. `tasks.py` schedules it every 120 s.

**Settings and API.** PEC fields on `sdi_it_settings` (`sdi_0002`): PEC address, SMTP host/port (465 SSL or STARTTLS), username, password (Fernet-encrypted, write-only, `has_smtp_password` in the response), IMAP host/port/folder, `sdi_pec_address` (validated to `*@pec.fatturapa.it`), `last_pec_poll_at`, `next_send_after`. Enabling with `transport = pec` requires complete credentials (400 otherwise). `POST /pec/test` logs in to both servers without sending; `POST /queue/process-now` runs one tick. Records carry `transport`, `message_id`, `next_attempt_at`. No new permission strings.

**The two follow-ups from #410's review.** `ProgressivoInvio` is bumped under a `SELECT … FOR UPDATE` on the settings row, so two invoices issued at the same instant cannot share a file name (test: two builds → `00001`, `00002`). Riepilogo `Imposta` is now base × rate rounded once (SDI check 00421); test: nine lines of 0.99 at 22 % → 8.91 / 1.96, where the old per-line sum gave 1.98.

**Tests** (36 + round-trip): fake SMTP/IMAP (message shape, Message-ID, attachment and zip extraction, seen-flagging, learning the SDI reply address, error wrapping), worker (send + apply, backoff and clinic pause, inert when manual/disabled, unmatched receipts counted not fatal), router (PEC settings gate, write-only password, `sdi_pec_address` domain guard, test endpoint mocked, process-now end to end), the Imposta and progressivo cases above. The round-trip test now downgrades both branch revisions (`sdi_it@-2`).

**Docs.** `docs/modules/sdi_it.md` (PEC section), `docs/technical/sdi_it/{overview,permissions}.md`, module `CHANGELOG.md`.

### DIFF
--- a/backend/app/modules/sdi_it/CHANGELOG.md
+++ b/backend/app/modules/sdi_it/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog — sdi_it
 
+## Unreleased — phase 2
+
+- PEC transport: the clinic's PEC mailbox sends each FPR12 file to the SDI
+  (SMTP) and a worker picks the receipts up (IMAP) every two minutes,
+  learning the dedicated `sdiNN@pec.fatturapa.it` reply address; retry with
+  backoff per file, ten-minute pause of the clinic on mailbox failures.
+- Settings: PEC fields (password encrypted at rest, write-only), `POST
+  /pec/test` (SMTP + IMAP login), `POST /queue/process-now`.
+- `Imposta` per riepilogo is now base × rate rounded once (SDI check 00421).
+- `ProgressivoInvio` counter bumped under a row lock, so two invoices issued
+  at the same instant cannot share a file name.
+- Records carry `transport` and `message_id`; receipt application shared
+  between the API import and the poller.
+
 ## 0.1.0 (2026-09-07) — phase 1
 
 - FatturaPA `FPR12` builder (TD01/TD04, `Natura N4` exempt lines, bollo
--- a/backend/app/modules/sdi_it/__init__.py
+++ b/backend/app/modules/sdi_it/__init__.py
@@ -7,13 +7,15 @@ SDI**. Invoices to natural persons for healthcare services may not be
 electronic (art. 10-bis DL 119/2018); they stay analogue and are reported
 to the Sistema Tessera Sanitaria by a separate module (ADR 0026).
 
-Phase 1 (this module): FPR12 XML for TD01/TD04, manual transport
-(download the file, upload it to the SDI, import the receipt), receipt
-handling RC/NS/MC. PEC transport and the Nuxt layer follow in their own
-PRs. See ``CLAUDE.md`` and ``docs/modules/sdi_it.md``.
+Phase 1: FPR12 XML for TD01/TD04, manual transport (download the file,
+upload it to the SDI, import the receipt), receipt handling RC/NS/MC.
+Phase 2: PEC transport — the clinic's PEC mailbox sends the files and a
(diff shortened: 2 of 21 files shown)


martinezsalmeron (approved):
Reviewed with `main` merged, sdi_it suite run in-container (38 green + round-trip deselected), ruff clean. Architecture is right: module-owned branch `sdi_0002` on `sdi_0001`, stdlib SMTP/IMAP in the executor with no new dependency, `ScheduledJob` through `get_scheduled_jobs()` so the worker obeys module state, receipt application shared between the API import and the poller, password through `app.core.email.encryption` like the notifications SMTP. The queue posture (lock → commit → network, per-row backoff, clinic pause) reads well and the tests exercise it with fakes rather than mocks of the worker itself. The Imposta rounding and the 400 on enabling PEC without credentials are both correct.

Two fixes pushed on top of the branch (06009b7d):

1. **Progressivo lock was ineffective.** `SELECT … FOR UPDATE` on a row already loaded in the session returns the identity-mapped instance *without* refreshing it, so the second concurrent request waited for the lock and then bumped the counter it had read before the first one committed — both invoices ended up as `IT…_00001.xml`. Reproduced with two sessions (new test in `test_hook.py`; the PR's own test runs both builds in one session, which is why it passed). `db.refresh(settings, with_for_update=True)` locks and reloads in one statement.
2. **PEC busta di trasporto.** A PEC provider never hands over the SDI's message as-is: it arrives as `From: "Per conto di: sdi07@pec.fatturapa.it" <posta-certificata@<gestore>>` with the original message attached as `postacert.eml`. `parseaddr` therefore returned the provider's mailbox, the "learn the dedicated `sdiNN@` address" branch never fired, and every later file kept going to `sdi01`. The receipt attachments themselves were already reachable (`walk()` descends into `message/rfc822`). `_sdi_sender` now finds the `pec.fatturapa.it` address in the outer From (display name included) or in the embedded message's From; test with an enveloped message.

Known ceiling, not blocking: messages are flagged `\Seen` before the receipts are applied, so a DB failure between poll and apply loses that batch from the poller (the receipt can still be imported by hand). Worth a note in the docs or a follow-up if it ever bites.

I could not smoke the endpoints over HTTP on my stack (sdi_it is not installed locally), the router tests cover them end to end.

PR #438, opened 2026-09-10
docs(adr): 0029 RBAC step 4 role_id writes and flag-aware callers (issue #46)

lamanji:
## Overview

- Splits ADR 0029 out of PR #415 per review (always-DB flip needs
  its own PR): RBAC step 4 role_id writes and flag-aware callers,
  with the SSE trade-off bullet corrected to stream-owned sessions.

## Tasks done

- `docs(adr)`: 0029 (split from #415 branch, bullet fixed)

## Modules touched

- None (docs only).

## Checklist

- [x] Docs layout gate green (gapless numbering verified)
- [x] No code, migration, or catalog changes

## Refs

- Split from #415 (issue #46, RBAC step 4).

## Review


ZoliQua (dismissed):
Read the ADR against the copy that was on #415's branch: identical except the SSE trade-off bullet, which now says what the code does after the review fixes (grants resolved in the stream factory on the stream-owned session). Structure follows `TEMPLATE.md`, status `proposed` with the always-DB flip left as an alternative pending Ramón's sign-off, and `check_docs_layout.py` is green; merges cleanly.

Two notes for the merge, neither blocking: (1) #439 also claims number 0029 — whichever lands second needs to renumber to 0030; (2) the "How to verify" paths (`tests/test_rbac_step4.py`, `rbac.py::granted_permissions_for`) exist only on #415, so this reads best merged together with or after #415. Approving from my side; `docs/adr/**` needs @martinezsalmeron either way.

martinezsalmeron (approved):
ADR content matches what #415 implements (role_id on every membership write, granted_permissions_for / has_permission_for as the two flag-aware chokepoints, always-DB flip left as a pending alternative). Renumbered to 0030 in bd2e3690: 0029 landed yesterday with the payment-gateway ADR (#445) and check_docs_layout rejects duplicate numbers. The 'How to verify' paths (test_rbac_step4.py, rbac.py helpers) arrive with #415; status stays proposed until the always-DB decision. Merging.

PR #439, opened 2026-09-10
Feat/payment gateways razorpay

tresundios:
## Summary
- Introduce a provider-neutral `payment_gateways` module with a shared adapter contract, registry, and async payment/refund lifecycle.
- Add the first provider implementation for Razorpay and wire it into the module activation flow.
- Add supporting migrations, permissions reuse, documentation, and tests for request/refund flows, idempotency, and uninstall safety.

## Modules touched
- payment_gateways
- razorpay
- payments
- frontend/deployment config

## Test plan
- Verify payment request creation, confirmation, cancellation, refresh, and refund flows.
- Verify duplicate webhook delivery stays idempotent.
- Verify module uninstall is blocked while requests or refunds are still in flight.
- Run backend and frontend lint/tests as applicable.


ZoliQua (changes requested):
Thank you — this is the shape #263 asked for: a neutral `payment_gateways` contract + registry, `PaymentRequest` / `GatewayRefundRequest` state machines, webhook confirmation under a row lock creating the `Payment` in-transaction, the `(provider, payment ref)` unique backstop, per-clinic Fernet-encrypted keys, httpx instead of the SDK, and a solid test set (signature valid/tampered/wrong secret, replay, amount mismatch, refund cap). Tenant filtering holds on every query I checked, including the webhook (clinic from the path selects the secret, `compare_digest`, then a locked lookup scoped to that clinic). `ruff`, eslint, docs layout and the catalogs are green locally and the branch merges cleanly onto main. Changes I need before approving:

1. **`pg_0001` must not chain off `pay_0004`.** With `branch_labels=("payment_gateways",)` and `down_revision="pay_0004"`, Alembic reports `payments@head` as ambiguous (`pg_0001` and `pay_0005` both descend from `pay_0004`), and `payment_gateways@base` walks down through payments → budget → odontogram → catalog → patients, so an uninstall downgrade would drop those tables. Use `down_revision="0001"` with `depends_on=("pay_0005", "pat_0003")` like `gdpr_0001` / `doc_0001`, and fix the "chained off `pay_0004`" sentences in CLAUDE.md/CHANGELOG (`pay_0005` has been the tip since #370). `rzp_0001` is already right.
2. **ADR:** the file is 0029 but eleven references still say "ADR 0022" (module CLAUDE.md files, docs/technical, glossary, user manual) — 0022 is the TOTP ADR. #438 also claims 0029; coordinate on 0030 if it lands first. Please set `Status: proposed` with the lead as decider, and `author` in both manifests to your handle.
3. **`payments.create.modal`** — a design question for @martinezsalmeron: the #263 review asked for the gateway's own collect panel via `payments.collect.actions` (#370). The PR instead adds a new slot that swaps the whole `PaymentCreateModal` for IN clinics with a 900-line copy of it: the `netbanking` chip #370 added is gone, the base modal's logic is duplicated, and a second provider could not register on the same slot ("first entry wins"). I'd rather have Ramón's call before you rework it.
4. `confirm()` should reject a `currency` mismatch the way it does for amount (the adapter defaults it to `INR`); `create_request` passes `ctx.clinic.currency` with no support check.
5. Please split out the unrelated changes: `Caddyfile.demo-in`, `docker-compose.demo-in.yml`, `docker-entrypoint.sh`, `docker-compose.yml`, `telephony/service.py`, `test_catalog_generator.py`, the trailing-newline change in `package.json` / `package-lock.json`, and the payments `ta.json` label.
6. Add `ar.json` to the razorpay layer and list it in `nuxt.config.ts` (the other nine are in perfect parity; CLAUDE.md's "fr/pt/… are a follow-up" is stale).
7. Webhook: `record_webhook_error` is rolled back by `get_db()` on the 401/500 paths — commit before raising; `RazorpayAdapter.verify_webhook_signature` is unused (the router uses the service's copy) — keep one.
8. `POST /requests/{id}/refresh` confirms without `FOR UPDATE`; a concurrent webhook loses on the unique index as a 500.

Smaller: per-row `gateway-info` calls on `/payments` (batch it); `modal.ondismiss` and a polling cap in the checkout flow; CSP entries for `checkout.razorpay.com` (`script-src`/`frame-src`/`connect-src` under enforce mode); `payment_date = captured_at.date()` is UTC; a test for the uninstall in-flight guard and a branch-scoped uninstall round-trip; the user-manual index still describes the "button next to New payment" design.

On #373: this PR implements the #263 plan, whereas #373 is a single-provider `/verify` flow without a webhook and currently conflicts with main. How the two relate is Ramón's decision; I've noted the comparison for him. `docs/adr/**`, `alembic.ini` and `permissions.ts` are CODEOWNERS paths.

Issue #446, opened 2026-09-11
Pagination and filtering in /settings/modules


BabuBahir:
### What problem does this solve

As the number of modules are increasing , there should be some way to filter installed , uninstalled , error prone  etc 

<img width="1820" height="991" alt="Image" src="https://github.com/user-attachments/assets/cf31bed2-15d3-4f7a-a5a5-a733540f7a7d" />

### What you have in mind

add filters and pagination  to /settings/modules

<img width="1767" height="732" alt="Image" src="https://github.com/user-attachments/assets/3b4d1545-3750-4e45-8cac-3b28175a8f7a" />

martinezsalmeron:
Shipped in #447 (1ec026b5): text search, state filter chip and 20-per-page pagination on /settings/modules, all URL-synced.

PR #427, opened 2026-09-09
feat(operations): module manager upgrade/installable signals + UI + docs (issue #47)

lamanji:
## Overview

- Follow-up to #54 (which delivered the `/settings/modules` page): the
  backend now exposes per-module `installable` + `upgrade_available`
  signals plus `installed_version`, the UI wires them (Upgrade button
  with "installed → on-disk" version, translated states, short
  dates in the detail modal), and the operations manual documents the
  module manager screens in English and Spanish. `installable` is a
  manifest passthrough; dependency reachability is enforced by
  Install, not by this signal.
- Rebuilt M5-clean off current main from the `pr/47-module-manager-phase1`
  commits; the i18n conflict was resolved by keeping main's shared
  plural rules and adding the named date format for all 10 locales
  (including Arabic).

## Tasks done

- `fix(core)`: keep installed version on reconcile; expose
  `upgrade_available` + `installable` + `installed_version` on
  `ModuleInfo`; only the upgrade finalize advances the stored
  version (issue #47)
- `feat(frontend)`: wire signals into ModuleCard/DetailModal, translated
  states, `d(value, 'short')` format (issue #47)
- `docs(operations)`: module manager UI section en+es (issue #47)

## Modules touched

- Core only (`core/plugins` + host frontend + operations manual). No
  manifest, migration, permission, or event changes.

## Checklist

- [x] PR title follows Conventional Commits
- [x] `eslint` clean on touched frontend files
- [x] Container pytest: `test_module_install_flow.py` +
  `test_module_service.py` 20/20 green
- [x] docs-layout OK, catalog check clean, no generated-file dirt
- [x] Reviewer pass: 0 blockers
- [x] M5-clean: merge-base is current `upstream/main`, only #47 files

## Test plan

- Open `/settings/modules` as admin: installable modules show install,
  outdated ones show an Upgrade button with "installed → on-disk",
  detail modal shows short dates
  with no console warnings in any of the 10 locales.
- `upgrade_available` is true only for installed modules whose manifest
  version differs from the last-applied record version.

## Refs

- Refs #47 (phase-1 follow-up; does not close the marketplace roadmap).

## Review

`@martinezsalmeron`: ready for review/merge when convenient.

PR #447, opened 2026-09-11
added filtering and pagination to settings

BabuBahir:
adds pagination & filtering to settings module page , for #446 446

Pagination length PAGE_SIZE = 20 , can be changed


ZoliQua (changes requested):
Thanks for picking up #446 — this is the right shape for it, and reusing `FilterBar` / `SearchBar` / `FilterChipMulti` / `PaginationBar` instead of inventing new controls is exactly what I hoped to see. Client-side filtering is the right call here too: the module list is a bounded registry the page already fetches whole, so there is no case for a new endpoint. I test-merged against current `main` (which now has #427) and there are no conflicts; the upgrade badge and the card survive intact.

Two things fail CI as it stands:

1. `settings.modules.searchPlaceholder` and `settings.modules.filterState` were added only to `en.json` and `es.json`. `tests/i18n/locale-parity.test.ts` requires every host locale to match `en.json` exactly, so `ar, de, fr, hu, it, pl, pt, ta` all fail. Please add both keys to the other eight files, Arabic included.
2. `tests/e2e/settings-modules.spec.ts` asserts the `patients` card is visible right after the page loads. The backend returns modules alphabetically and we ship around fifty, so `patients` lands near position 29 — page 2 at `PAGE_SIZE = 20`. Either search for it in the spec first, or assert on a module inside the first twenty.

Two behaviour issues:

3. Browser Back loses the page number. In the `route.query` watcher, assigning `search` triggers `watch(search)`, which resets `page` to 1 after the watcher has restored it from the URL, so `?q=pay&page=2` comes back as page 1. A `syncingFromUrl` flag that the `search`/`stateFilter` watchers check before resetting `page` fixes it.
4. `pushUrl` rebuilds the query from scratch, so any unrelated query parameter is dropped on the first filter change. Spread `...route.query` and delete only the three keys this page owns.

And a docs item: #427 landed `operations.md` §13 "Module manager UI" on `main` today, and its module-list bullet now describes the page without your filter bar. Please rebase and add a line in both `docs/user-manual/en/operations.md` and the Spanish one covering the search box, the state filter, twenty per page, and the URL-synced `?q=&states=&page=` — a linkable filtered view is a detail worth documenting.

Not blocking: `SearchBar` already has a `debounce` prop and an `@update:debounced` event, so the second 300 ms timer in the page is redundant; `STATE_I18N_KEY` duplicates `camelState` in `ModuleCard.vue`, worth lifting into `~/utils` and importing in both; and a `scrollTo(0, 0)` on page change would stop Next from dropping the reader mid-list.

Once these are in it needs @martinezsalmeron's approval as well — `frontend/app/**` is a CODEOWNERS path.

ZoliQua (changes requested):
Good round — the two CI blockers are gone and the URL handling is right now. Verified on `2eef8dde`:

- `settings.modules.searchPlaceholder` and `settings.modules.filterState` are in all ten host locales.
- `isSyncingFromUrl` guards the `search` and `stateFilter` watchers, so a Back after paging no longer collapses to page 1.
- `pushUrl` and `pushPage` both start from `...route.query`, so unrelated query parameters survive.
- The CHANGELOG entry reads well.

Three things left, one of which will bite CI:

1. **The e2e placeholder is English-only.** `getByPlaceholder(/search modules/i)` sits two lines under a heading check written as `/módulos|modules/i`, which is there precisely because the run's language is not fixed. In Spanish the placeholder is "Buscar módulos…" and the fill throws. Please make it `/buscar módulos|search modules/i` like the line above it.
2. **The branch history carries a stray file.** `2eef8dde` deletes `backend/D:/PROJECTS/GitHUbCOntris/dentalpin/frontend/modules.json`, which an earlier commit had added — a 155-line blob committed under a Windows absolute path. The final tree is clean, but we merge with merge commits, so it would stay in the repository's history for good. Please squash the branch (or rebase it out) before this goes in.
3. **The docs ask from the last round is still open.** #427 landed `operations.md` §13 "Module manager UI" on `main`, and its module-list bullet describes the page without your filter bar. One line in both `docs/user-manual/en/operations.md` and the Spanish one: search box, state filter, twenty per page, and the URL-synced `?q=&states=&page=` so a filtered view can be linked.

Two nits, take them or leave them: `pushUrl` and `pushPage` are now byte-identical apart from `router.replace` versus `router.push`, so one function with a `mode` argument would do; and `STATE_I18N_KEY` still duplicates `camelState` in `ModuleCard.vue`, worth lifting into `~/utils`. The `/.vs` line in `.gitignore` is unrelated but harmless, no need to remove it.

After these it needs @martinezsalmeron as well, since `frontend/app/**` is a CODEOWNERS path.

ZoliQua (changes requested):
Two of the three are done and the branch is much cleaner:

- The e2e now fills `getByPlaceholder(/buscar módulos|search modules/i)`, matching the bilingual heading assertion above it.
- The branch is squashed to two commits and the stray `backend/D:/PROJECTS/GitHUbCOntris/…/modules.json` blob is gone from its history.
- Re-checked what was already good: both keys in all ten locales, `isSyncingFromUrl` guarding the `search`/`stateFilter` watchers, and `{ ...route.query }` in both push helpers.

The third is still open, and it is the one I cannot waive, because it is in the contributing bar: **the operations manual**. #427 landed §13 "Module manager UI (`/settings/modules`)" on `main`, and its module-list bullet describes the page as it was before your filter bar. One line in `docs/user-manual/en/operations.md` and the matching one in `docs/user-manual/es/operations.md`: the search box, the state filter, twenty per page, and that the view is URL-synced as `?q=&states=&page=` so a filtered list can be linked or bookmarked. That last detail is genuinely useful to an operator and nobody will discover it from the UI.

I also approved the CI run, which had been sitting in "action required" — it is running now. Once the docs line is in and CI is green this needs @martinezsalmeron for `frontend/app/**`, and I would expect it to go through quickly.

ZoliQua (dismissed):
Correction, and an apology: the documentation **is** in this PR and always was. My last review said it was missing — that was my mistake, not yours. I checked with a two-dot `git diff main..branch`, which compares the tip of `main` against your branch rather than showing what the branch changed since it diverged. The three-dot form shows it plainly:

```
docs/user-manual/en/operations.md
docs/user-manual/es/operations.md
```

and the §13 entry reads exactly as asked — search box, state filter, twenty per page, and the `?q=`, `?states=`, `?page=` URL sync with a note that a filtered view can be bookmarked or linked. The Spanish one matches. Sorry for sending you back to something already done.

So everything from all three rounds is in: both i18n keys in all ten locales, the bilingual `getByPlaceholder` in the e2e, `isSyncingFromUrl` guarding the watchers, `{ ...route.query }` in both push helpers, the squashed history with the stray Windows-path blob gone, and the manual in both languages. CI is 12/12 green.

Approving. `frontend/app/**` is a CODEOWNERS path, so @martinezsalmeron's approval is the only thing left — nothing further needed from you.

Issue #444, opened 2026-09-11
Refreshing page after logging in logs out and redirects back to login page


BabuBahir:
### What happened

1. Login using default credentials https://demo.dentalpin.com/login
2.  Once you log in  navigate to https://demo.dentalpin.com/patients  or any other route like https://demo.dentalpin.com/payments
3. Just Refresh !!!!
4.  You are logged out !!

### Steps to reproduce

1. Login using default credentials https://demo.dentalpin.com/login
2.  Once you log in  navigate to https://demo.dentalpin.com/patients  or any other route like https://demo.dentalpin.com/payments
3. Just Refresh !!!!
4.  You are logged out !!

<img width="1907" height="1029" alt="Image" src="https://github.com/user-attachments/assets/51857a6d-260a-4fc3-83b9-62a656c57142" />

ZoliQua:
Reproduced, and it is a deployment setting on the demo rather than a bug in the app — thank you for the precise report, it is exactly the kind that is easy to act on.

The demo runs the app on `demo.dentalpin.com` and the API on `api-demo.dentalpin.com`. Since #394 the session lives in cookies that the API sets, and the login response sets them without a `Domain` attribute:

```
set-cookie: dp_access=…; HttpOnly; Max-Age=900; Path=/; SameSite=lax; Secure
set-cookie: dp_refresh=…; HttpOnly; Max-Age=604800; Path=/; SameSite=lax; Secure
set-cookie: dp_csrf=…; Max-Age=604800; Path=/; SameSite=lax; Secure
```

A cookie with no `Domain` is host-only, so the browser sends those three to `api-demo.dentalpin.com` and never to `demo.dentalpin.com`. Client-side navigation therefore works (every call goes to the API host with `credentials: include`), but a refresh is server-rendered on the app host, where the incoming request carries no session cookie at all, so the auth middleware does the only thing it can and redirects to `/login`.

Same request, twice, against the live demo:

```
# what your browser does today: no cookies reach the app host
GET https://demo.dentalpin.com/patients          → 302 → /login

# the same with the three cookies present on the app host
GET https://demo.dentalpin.com/patients          → 200
```

The fix is the `COOKIE_DOMAIN` setting the split-host case was built for (`backend/app/config.py`, `.env.example`, and it is already passed through in `docker-compose.coolify.yml`): set

```
COOKIE_DOMAIN=.dentalpin.com
```

[...]

Issue #448, opened 2026-09-11
Uninstalling/installing a module throws error in demo site


BabuBahir:
### What happened

Uninstalling/installing a module throws error in demo site https://demo.dentalpin.com/settings/modules

<img width="1506" height="858" alt="Image" src="https://github.com/user-attachments/assets/2032e3a0-59b1-40da-9e74-1f7cee1c6df1" />

### Steps to reproduce

1. login to https://demo.dentalpin.com/settings/modules
2.  Uninstall / Install any module
3.  Get Error

 

### Relevant logs

```shell
Request URL :https://api-demo.dentalpin.com/api/v1/modules/recalls/uninstall
Request Method :POST
Status Code :403

{
    "data": null,
    "message": "CSRF token missing or invalid",
    "errors": [
        "CSRF token missing or invalid"
    ]
}
```

ZoliQua:
Same root cause as #444, and worth stating plainly: **every write on the demo fails right now, not just module install and uninstall.**

The demo serves the app from `demo.dentalpin.com` and the API from `api-demo.dentalpin.com`, and the API sets the session cookies without a `Domain` attribute, so all three are host-only for the API host. I re-checked it just now:

```
set-cookie: dp_access=…;  HttpOnly; Max-Age=900;    Path=/; SameSite=lax; Secure
set-cookie: dp_refresh=…; HttpOnly; Max-Age=604800; Path=/; SameSite=lax; Secure
set-cookie: dp_csrf=…;              Max-Age=604800; Path=/; SameSite=lax; Secure
```

`dp_csrf` is the JS-readable half of the double-submit pair. The app reads it with `useCookie('dp_csrf')` on its own origin (`frontend/app/composables/useSessionRequest.ts`) and echoes it in `X-CSRF-Token` on every unsafe method. On `demo.dentalpin.com` that cookie does not exist, so the header is omitted and the backend correctly answers `403 CSRF token missing or invalid` — for uninstall, install, and equally for saving a patient, issuing an invoice or changing a setting. Reads work because the browser attaches the cookies to the API host by itself, which is why the demo looks healthy until you try to change something.

The fix is the same single line as #444: set

```
COOKIE_DOMAIN=.dentalpin.com
```

in the demo's environment and redeploy. Both hosts are under that parent, so the browser will send the cookies to the app host too; `dp_csrf` becomes readable there and the header is sent again.

[...]

PR #449, opened 2026-09-11
fix(auth): warn when session cookies can never reach a split-host app (#444)

ZoliQua:
## Overview

#444 (thanks @BabuBahir) is a deployment gap, not an app bug, but it is one nothing in the repository warns about, so it is worth a guard.

The demo serves the app from `demo.dentalpin.com` and the API from `api-demo.dentalpin.com`, and the login response sets the session cookies without a `Domain`:

```
set-cookie: dp_access=…; HttpOnly; Max-Age=900; Path=/; SameSite=lax; Secure
```

A cookie without `Domain` is host-only, so the browser sends the three cookies to the API host and never to the app host. Client-side navigation works; a reload is server-rendered on the app host with no session cookie at all, and the auth middleware redirects to `/login`. Against the live demo:

```
GET https://demo.dentalpin.com/patients                      → 302 → /login
GET https://demo.dentalpin.com/patients  (cookies present)   → 200
```

`COOKIE_DOMAIN` (#394) exists for exactly this and is empty on the demo; setting `COOKIE_DOMAIN=.dentalpin.com` there fixes it. The failure mode is silent at deploy time and only shows up on reload, so:

- `warn_if_host_only()` in `backend/app/core/auth/cookies.py` logs **once per process**, the first time session cookies are issued, when the request's `Origin` (or `Referer`) host differs from the API's own host and `COOKIE_DOMAIN` is empty. It names both hosts and the exact value to set, derived from the longest common suffix of the two hosts (`.dentalpin.com`), or says to put them under one parent domain when there is no shared one.
- `set_session_cookies()` takes the request and calls it, so login and refresh are both covered and it cannot be forgotten at a call site.
- `docs/user-manual/{en,es}/operations.md`: a prerequisite bullet for split-host deployments and a troubleshooting row ("logged out on every page refresh, while clicking around works").

No behaviour change otherwise: no new setting, nothing fails, single-host deployments log nothing.

## Test plan

- `backend/tests/test_auth_cookies.py`: split-host login logs the warning containing `COOKIE_DOMAIN=.example.com`; no warning when both hosts match or when `COOKIE_DOMAIN` is already set; `_shared_parent` unit test (needs two common labels, so `app.com` + `api.com` yields nothing).
- `pytest tests/test_auth_cookies.py` → 14 passed. `ruff check` / `format --check` clean. `check_docs_layout.py` OK.
- Verified by hand against the live demo (both curl lines above).

The deployment side of #444 stays with you: `COOKIE_DOMAIN=.dentalpin.com` on the demo.

Refs #444

### DIFF
--- a/backend/app/core/auth/cookies.py
+++ b/backend/app/core/auth/cookies.py
@@ -63,10 +63,14 @@ def warn_if_host_only(request: Request | None) -> None:
 
     A cookie without ``Domain`` is host-only. When the app is served from
     another host than the API, the browser keeps the session cookies for
-    the API host: client-side calls still work, but a server-rendered page
-    load carries no cookie at all, so every reload lands on ``/login``.
-    Nothing fails loudly at deploy time, so say it here, once, with the
-    value to set.
+    the API host. Client-side reads still work, because the browser
+    attaches them to the API host itself, so the deployment looks healthy;
+    what breaks is everything that needs the cookies on the app's own
+    origin. A server-rendered page load carries no cookie at all and lands
+    on ``/login`` (#444), and ``dp_csrf`` - the JS-readable half of the
+    double-submit pair - is unreadable there, so every unsafe request goes
+    out without ``X-CSRF-Token`` and is refused with 403 (#448). Nothing
+    fails at deploy time, so say it here, once, with the value to set.
     """
     global _HOST_ONLY_WARNED
     if _HOST_ONLY_WARNED or _domain() or request is None:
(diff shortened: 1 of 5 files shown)


martinezsalmeron (approved):
Reviewed: 14/14 cookie tests green in the container, and exercised live on the dev stack — same-host login logs nothing, split-host login logs the warning exactly once with `set COOKIE_DOMAIN=.dentalpin.com and restart`. Confirmed against the live demo too: the login response still sets the three cookies without `Domain`, so #444 and #448 are both this deployment gap. Small, contained, log-only. Merging; the `COOKIE_DOMAIN` change on the demo is on our side.

Issue #332, opened 2026-08-29
feat(copilot): pluggable LLM providers - free & local tiers (design first)


lamanji:
The AI gateway currently ships OpenAI-only: `app/core/llm/factory.py` has `SUPPORTED_PROVIDERS = ("openai",)` and a factory built to take more (its own docstring: "Anthropic / Ollama slot in here later"). Design-first, aligning with roadmap copilot (#223):

- Provider registry: Anthropic, Google Gemini, OpenRouter, Groq (free tier), Ollama (local) alongside OpenAI.
- Per-clinic provider + model selection via settings (encrypted API key - same pattern as payroll #229's encrypted fields).
- Free-tier options for evaluation/demo without a paid key.
- Cost & runtime guardrails (token caps, per-role enable).
- No behavior change to the copilot UI/floating button.

Kept separate from the navigation reorg (#232) per discussion.

ZoliQua:
Anthropic adapter is up: #342 — fitted to the current factory (no registry/manifest opinions, that stays in the #340 design discussion). The `tool_to_anthropic_schema` serializer that already existed in the agents layer is now actually selected by the bridge when the clinic's provider is `anthropic`.

lamanji:
Status update with evidence: the core of this issue is delivered.

- #340 (design brief for pluggable providers, free and local tiers) merged 2026-08-31.
- #342 (Anthropic adapter fitted to the current factory, selected by the bridge when the clinic's provider is `anthropic`) merged 2026-08-31.

Still open from the body: Gemini / OpenRouter / Groq / Ollama adapters, per-clinic provider plus model selection in settings, and cost plus runtime guardrails. Suggest keeping this as the umbrella and tracking each adapter separately - point me at which one (or the settings selection) to take next and I will pick it up once the current review queue clears.

ZoliQua:
One piece missing from that list, and it matters for what gets picked up next: **#456** (@Aryan-coder-student) is the provider registry this issue asked for. `ProviderSpec` carries `default_model`, `tool_dialect`, `needs_api_key` and `api_key_setting`; copilot registers the built-ins from `on_activate()`; and the hardcoded provider branches in `copilot/service.py` and `bridge.py` are gone. I approved it (twice — the merge dismissed the first); it is waiting on @martinezsalmeron because `backend/app/core/llm/**` is a CODEOWNERS path.

[...]

Issue #452, opened 2026-09-12
auth: four non-blob fetch call sites still bypass the 401 refresh (media list/upload/delete/update, copilot stream)


ZoliQua:
Follow-up to #440 / #451, which moved every **blob** download onto `useApi().raw()` and gave that helper a refresh-and-retry on 401. Four call sites are left that do their own `fetch`/`$fetch` with `credentials: 'include'` and therefore still fail outright when the access cookie has expired.

**`media/useDocuments.ts`** — four `$fetch` calls: the document list, the upload, the delete and the update. They were left out of #451 on purpose: the upload carries `FormData` with `onRequestError`/`onResponse` progress callbacks, which `useApi` does not model, and the list/delete/update are JSON calls that should simply move to `api.get`/`api.del`/`api.put` rather than to `raw()`. Doing it properly means either teaching `useApi` about progress, or keeping the upload on `$fetch` and wrapping only the other three.

Symptom today: after an idle longer than `ACCESS_TOKEN_EXPIRE_MINUTES`, opening a patient's Documents tab shows an empty list with a console error instead of recovering, and an upload started from a stale tab fails once and works on the second try.

**`copilot/useCopilotStream.ts`** — the SSE stream opens with a raw `fetch`. A 401 here cannot be retried blindly: the stream may already have emitted tokens, so the fix is to refresh before opening when the response is 401 and the body has not started, and to surface a clear "session expired" state otherwise. Worth its own thought rather than a mechanical swap.

Not in scope: `budget/usePublicBudget.ts`, which is the public token flow and has no session to refresh.

**Suggested shape**

1. `media/useDocuments`: list, delete and update through `useApi`; keep the upload on `$fetch` but retry it once after `auth.refresh()` on a 401.
2. `copilot/useCopilotStream`: refresh once before the first chunk; otherwise end the stream with a session-expired message.
3. A test for each, mirroring `frontend/tests/composables/useApiRaw.test.ts`.

PR #456, opened 2026-09-12
refactor(copilot): add LLM provider registry (#332)

Aryan-coder-student:
## Summary

Refactor Copilot LLM provider resolution to use a provider registry instead of hardcoded OpenAI/Anthropic branches.

Existing OpenAI and Anthropic providers are registered from `CopilotModule.on_activate()` following ADR 0020. No new providers or settings behaviour are added in this PR.

Refs #332.

## Type of Change

* [ ] Bug fix
* [ ] New feature
* [x] Refactoring
* [ ] Documentation

## Testing

* `cd backend && ruff check . && ruff format --check .`
* `cd frontend && npm run lint`
* `cd backend && python -m pytest tests/test_llm_provider_registry.py -v`
* `cd backend && python -m pytest tests/test_llm_orchestrator.py -v`

Added tests for provider registration, lookup, re-registration, overrides, unregister, Copilot activation, and factory resolution.

##  Review
@martinezsalmeron : ready for review

### DIFF
--- a/backend/app/core/llm/factory.py
+++ b/backend/app/core/llm/factory.py
@@ -1,16 +1,46 @@
-"""Provider resolution.
-
-Resolves ``"openai"`` and ``"anthropic"``. Further vendors (Gemini,
-Ollama) slot in here later with no change to callers — the orchestrator
-already speaks neutral types (``base.py``).
-"""
+"""Provider resolution through the process-wide LLM provider registry."""
 
 from __future__ import annotations
 
 from app.config import settings
 from app.core.llm.base import LLMConfigError, Provider
+from app.core.llm.registry import llm_provider_registry
+from app.core.llm.spec import ProviderConfig, ProviderSpec
 
-SUPPORTED_PROVIDERS = ("openai", "anthropic")
+
+def _create_openai(config: ProviderConfig) -> Provider:
+    from app.core.llm.openai_provider import OpenAIProvider
+
+    return OpenAIProvider(api_key=config.api_key or settings.OPENAI_API_KEY)
+
+
+def _create_anthropic(config: ProviderConfig) -> Provider:
+    from app.core.llm.anthropic_provider import AnthropicProvider
+
+    return AnthropicProvider(api_key=config.api_key or settings.ANTHROPIC_API_KEY)
+
+
+OPENAI_SPEC = ProviderSpec(
+    name="openai",
+    label="OpenAI",
+    tier="paid",
+    default_model=settings.COPILOT_MODEL_CHAT_OPENAI,
+    supports_tools=True,
+    redaction_required=True,
+    needs_api_key=True,
+    factory=_create_openai,
+)
+
+ANTHROPIC_SPEC = ProviderSpec(
+    name="anthropic",
+    label="Anthropic",
+    tier="paid",
+    default_model=settings.COPILOT_MODEL_CHAT_ANTHROPIC,
+    supports_tools=True,
+    redaction_required=True,
+    needs_api_key=True,
+    factory=_create_anthropic,
+)
 
 
 def get_provider(name: str, *, api_key: str | None = None) -> Provider:
@@ -19,16 +49,11 @@ def get_provider(name: str, *, api_key: str | None = None) -> Provider:
     Raises :class:`LLMConfigError` for unsupported names so a clinic can
     never select a provider this deployment cannot serve.
     """
-    if name == "openai":
-        from app.core.llm.openai_provider import OpenAIProvider
+    spec = llm_provider_registry.get(name)
+    if spec is None:
+        supported = ", ".join(item.name for item in llm_provider_registry.list()) or "none"
+        raise LLMConfigError(
+            f"Unsupported LLM provider: {name!r} (supported: {supported})"
+        )
 
-        return OpenAIProvider(api_key=api_key or settings.OPENAI_API_KEY)
-
-    if name == "anthropic":
-        from app.core.llm.anthropic_provider import AnthropicProvider
-
-        return AnthropicProvider(api_key=api_key or settings.ANTHROPIC_API_KEY)
-
-    raise LLMConfigError(
-        f"Unsupported LLM provider: {name!r} (supported: {', '.join(SUPPORTED_PROVIDERS)})"
-    )
+    return spec.factory(ProviderConfig(api_key=api_key))
--- a/backend/app/core/llm/factory.py
+++ b/backend/app/core/llm/factory.py
@@ -52,8 +52,6 @@ def get_provider(name: str, *, api_key: str | None = None) -> Provider:
     spec = llm_provider_registry.get(name)
     if spec is None:
         supported = ", ".join(item.name for item in llm_provider_registry.list()) or "none"
-        raise LLMConfigError(
-            f"Unsupported LLM provider: {name!r} (supported: {supported})"
-        )
+        raise LLMConfigError(f"Unsupported LLM provider: {name!r} (supported: {supported})")
 
     return spec.factory(ProviderConfig(api_key=api_key))
--- a/backend/app/core/llm/registry.py
+++ b/backend/app/core/llm/registry.py
@@ -1,0 +1,40 @@
+"""Process-wide registry for pluggable LLM providers."""
+
+from __future__ import annotations
+
+import logging
+
+from app.core.llm.spec import ProviderSpec
+
+logger = logging.getLogger(__name__)
+
+
+class ProviderRegistry:
+    """Map provider names to active provider specifications."""
+
+    def __init__(self) -> None:
+        self._providers: dict[str, ProviderSpec] = {}
+
+    def register(self, spec: ProviderSpec) -> None:
+        """Register ``spec`` by name, idempotently replacing overrides."""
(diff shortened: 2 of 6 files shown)


ZoliQua (changes requested):
Thanks for this, and welcome — the registry itself is the right shape and the tests are unusually careful (a local `ProviderRegistry()` for the unit cases, `finally` cleanup around the global one, idempotent re-registration covered). I checked the refactor cannot break anything on the way in: `SUPPORTED_PROVIDERS` had no importer outside `factory.py`, and the only caller of `get_provider` is copilot's own bridge, which runs well after `on_activate()`. Registering from `on_activate()` is exactly what ADR 0020 asks for. Locally: `ruff check` and `ruff format --check` clean, and `test_llm_provider_registry.py` plus every copilot and LLM test file is green (47 passed). I approved the CI run, which was waiting for a maintainer.

What I would like changed before this goes in:

1. **The hardcoded provider branches are still there, one layer up.** `copilot/service.py` keeps `_DEFAULT_MODEL_BY_PROVIDER = {"openai": …, "anthropic": …}` and two `if data.get("provider") == "…" and not app_settings.…_API_KEY` checks, and `bridge.py:121` still reads `return "anthropic" if provider_name == "anthropic" else "openai"` for the tool-schema dialect. So after this PR, adding Gemini or Ollama still means editing three hardcoded places — which is the thing #332 exists to end. Please have those three read the registry: `spec.default_model` for the fallback, `spec.needs_api_key` (plus a way to name the settings key) for the validation, and a `tool_dialect` field on the spec for the bridge.
2. **Half the spec is metadata nothing reads.** `label`, `tier`, `supports_tools`, `redaction_required` and `default_model` have no consumer in this PR, and `default_model` and `needs_api_key` currently duplicate values that live in `service.py`. Two sources of truth drift quietly. Either wire them as in point 1, or ship only the fields the registry uses today and add the rest with the code that reads them.
3. **The architecture doc is now wrong.** `docs/technical/copilot-agentic-architecture.md` line 43 still says `get_provider` resolves "`openai`" and "`anthropic`" with "further vendors slot in later"; that paragraph should describe the registry and where registration happens. The docstring in `core/llm/__init__.py` still says "v1: `openai`" too. Docs belong with the change in this repo.

Small one, not blocking: `test_copilot_activation_registers_builtin_providers` unregisters `openai` and `anthropic` in its `finally`, so the global registry ends the session without the built-ins. Nothing depends on that today, but a later test that asserts `get_provider("openai")` would fail depending on collection order — re-registering the two specs (or using `monkeypatch`) keeps the process state as it found it.

One design note for @martinezsalmeron rather than for you: the specs are defined in `core/llm/factory.py` and registered by the `copilot` module, so core still carries the built-in provider knowledge and the registry only pays off for providers a *different* module contributes. That reads fine to me and matches how `sms_gateway` does it, but `backend/app/core/llm/**` is a CODEOWNERS path, so his approval is required either way.

Aryan-coder-student:
Here are the changes as requested @ZoliQua @martinezsalmeron 

- Removed hardcoded model and API-key branches; Copilot now reads provider metadata from the registry (backend/app/modules/copilot/service.py:L32, L69-L79).

- Tool dialect now comes from the provider specification (backend/app/modules/copilot/bridge.py:L119-L121).

- Added generic credential resolution through api_key_setting (backend/app/core/llm/factory.py:L53-L57).

- Removed unused metadata and added spec validation (backend/app/core/llm/spec.py:L20-L36).

- Updated the architecture documentation (docs/technical/copilot-agentic-architecture.md:L44-L46).

- Fixed activation-test cleanup to restore registry state (backend/tests/test_llm_provider_registry.py:L88-L110).

ZoliQua (dismissed):
All four points are addressed, and addressed the way I hoped rather than the cheapest way. Thanks for the careful pass — and sorry for the slow turnaround.

What I verified on `d861f086`:

1. **No provider branches left.** `service.py` has no `_DEFAULT_MODEL_BY_PROVIDER` and no `== "openai"` / `== "anthropic"` checks; the key validation now reads `spec.needs_api_key` and names the setting via `spec.api_key_setting`, and the model fallback is `spec.default_model`. `bridge.py:121` is `get_provider_spec(provider_name).tool_dialect`. I grepped the whole backend for `get_provider`/`get_provider_spec`/`llm_provider_registry`: every consumer is inside `copilot`, and there is no remaining provider-name comparison anywhere.
2. **The spec is only what is read.** `tier`, `supports_tools` and `redaction_required` are gone; `label` now earns its place in the settings error message. `default_model` reads from `settings.COPILOT_MODEL_CHAT_*` in the spec itself, so the duplication is gone — one source of truth.
3. **Docs match the code.** `copilot-agentic-architecture.md` describes the registry and names `on_activate()` as the registration point; the `core/llm/__init__.py` docstring no longer says "v1: openai".
4. **The test restores the process state.** `test_copilot_activation_registers_builtin_providers` snapshots both specs and re-registers them in `finally`. Good.

Also checked: `on_activate()` is documented in `BaseModule` as running once per boot for installed modules, after mount — so the registry is populated before any copilot route can be reached, and a restart re-attaches the built-ins. Nothing resolves a provider before that point.

Locally, on your head commit: `ruff check` and `ruff format --check` clean; `test_llm_provider_registry.py` + `test_copilot_bridge.py` + `test_copilot_digest.py` = 27 passed; the full suite is 1046 passed with only failures that also fail on `main` on my machine (WeasyPrint has no libpango here). CI is green.

**Approving.** Two things before it can land:

* **The branch conflicts with `main`** — one hunk, `backend/app/modules/copilot/CHANGELOG.md`, where #467 and #452 landed entries under `## Unreleased` next to yours. Keep both sides. Please rebase or merge `main` in.
* `backend/app/core/llm/**` is a CODEOWNERS path, so @martinezsalmeron's approval is required on top of mine.

Two non-blocking notes, either for now or for a follow-up:

* `get_configured_api_key` does `getattr(settings, spec.api_key_setting)` with no default. For the built-ins that is always fine, but a module registering a provider whose key setting is not a field on the core `Settings` gets an `AttributeError` from an `os`-level typo instead of the clean "… is not configured" `ValueError` two lines later. `getattr(settings, spec.api_key_setting, None)` makes it degrade into the message you already wrote.
* `get_or_create` now raises `LLMConfigError` when `COPILOT_PROVIDER_DEFAULT` names an unregistered provider, where before it quietly created a settings row and failed later at send time. I think that is an improvement — failing at config read with a list of supported names beats failing mid-conversation — just flagging it as a deliberate behaviour change so nobody is surprised.

@martinezsalmeron — the design note from my first review still stands: the built-in specs live in `core/llm/factory.py` and are registered by `copilot`, so core still carries the built-in provider knowledge and the registry only pays off for a provider some *other* module contributes. The `getattr` point above is the same seam from the other side: a third-party provider's API key still has to be a field on core `Settings`. That reads fine to me for #332's scope, but it is your call.

ZoliQua (approved):
Re-approving after the merge — GitHub dismissed my approval when you pushed `35329b4b`.

I diffed the PR's own contribution before and after: `main...head` is byte-identical to the tree I approved (same 12 files, 385 insertions / 58 deletions, and the `+`/`-` lines match exactly), so the merge only resolved the `copilot/CHANGELOG.md` conflict, and it kept both sides — the #232 and #452 entries are still there next to yours. The branch is mergeable now.

Re-ran on the new head against current `main` (which has the Sentry work in it): `test_llm_provider_registry.py` + `test_copilot_bridge.py` + `test_copilot_digest.py` + `test_error_tracking.py` = 32 passed. CI green.

Still needs @martinezsalmeron for `backend/app/core/llm/**` (CODEOWNERS). The two non-blocking notes from my last review stand as written — neither needs to hold this up.

martinezsalmeron (approved):
Approving for `backend/app/core/llm/**` (CODEOWNERS). Thanks @Aryan-coder-student — clean first contribution, and thanks @ZoliQua for the two careful rounds.

What I checked on `35329b4b`:

- **Architecture.** `ProviderSpec`/`ProviderRegistry` live in core, core imports no module, and copilot registers the built-ins from `on_activate()` — same ADR 0020 pattern as notifications and payment_gateways. No process resolves against an empty registry: `mount_modules` runs `on_activate()` before the module goes live, the in-process scheduler starts afterwards, and the digest tasks don't go through `get_or_create`.
- **Function.** No provider-name branches left in `service.py` / `bridge.py`. It also fixes two latent bugs on `main`: an unknown `provider` in the settings PATCH used to be persisted silently (now rejected), and `get_or_create` seeded the OpenAI model even when `COPILOT_PROVIDER_DEFAULT` was `anthropic`.
- **Tested.** Real chat in the browser (tool calls + streaming) and the rejecting PATCH via API; full backend suite green locally and in CI on this head. `main` has moved since, but nothing merged touches any of these 12 files.

Non-blocking, fine as a follow-up (or with the first new provider for #332):

1. `ProviderConfig.base_url` has no consumer yet — drop it or wire it when Ollama/compatible endpoints land.
2. `getattr(settings, spec.api_key_setting)` has no default, so a typo'd setting name surfaces as `AttributeError` at request time; validating it in `ProviderSpec.__post_init__` or at `register()` would fail at boot instead.
3. `backend/app/modules/copilot/CLAUDE.md` doesn't mention the provider registry / `on_activate()` registration.

PR #460, opened 2026-09-14
Feat/mcp init

BabuBahir:
This is a Foundation for MCP abilities for Dentalpin.
This MCP feature is added as a module , as of now it exposes abilities/tools  (getPatient(gets 1 patient) , SearchPatients , CreatePatient).

If enabled the mcp is exposed at http://[Some host]:8000/mcp.   
Agents like claude/kilo and other AI agents can then use these tools to operate their dentalpin.

The authentication of mcp is done using a dt_token which is derived from an access_token .
I have also added a claude skill for agents to obtain the token and also tutorial on how to obtain the dt_ token manually via postman 

The mcp client is then connected . And operations can be done .

Think of it like a DentalPin as a service Where a clinic imports/exports data via Agents and not via traditional API/UI .

I am going to attach more screenshots of how I tested it locally


ZoliQua:
Thanks for putting this up as a draft and asking for views before polishing — that was the right call, and the core design is better than I expected. Routing every call through `tool_registry.call` instead of building a second data path is exactly right: RBAC, guardrails, validation and the audit log all stay at one chokepoint, `clinic_id` comes off the token rather than the request, and translating token scopes into RBAC grants so a read-only token cannot even see `create_patient` in `tools/list` is a genuinely nice piece of design. The documentation and the ADR are thorough.

There is one thing I cannot decide and should not: `server.py` hardcodes `AgentMode.AUTONOMOUS`, so a `patients:write` token held by a third-party AI host creates patient records with no human in the loop, and the mode is not configurable per token. Whether DentalPin grants that, and at what granularity, is a product decision for @martinezsalmeron. My own instinct is that it should land `SUPERVISED` with autonomous behind an explicit per-token opt-in, but it is his call. Related and deliberate-looking: the `.gitignore` change un-ignores all of `.claude/` rather than just the skills directory, so every contributor's local `.claude/` tree — worktrees included — becomes committable. That is a repo-wide policy change and should be decided on purpose, not as a side effect of shipping the token-minting skill.

Mechanically, before this leaves draft:

- `backend/D:/PROJECTS/GitHUbCOntris/dentalpin/frontend/modules.json` is committed — a Windows absolute path pasted under `backend/`. The same stray file turned up in #447's history, so it looks like something in your setup writes it; worth finding the source. Please remove it here. `.kilo/kilo.jsonc` and the `.vs` ignore rules are editor tooling for one setup and belong in their own PR if anywhere.
- `uv.lock` was regenerated with an older `uv`: it drops `revision = 3`, `provides-extras = ["dev"]` and the `upload-time` field on about 1550 wheels. No package or version actually changed — eleven added, none removed — but the lockfile-drift job will fail. Re-run `uv lock` with the repo's pinned version so the diff is only the new packages.
- `pyproject.toml` pins `mcp>=1.0` while the code uses the 2.x lowlevel API, so a fresh install without the lock breaks on import. Please pin `>=2.2,<3`. The server also imports `mcp_types` and the test imports `httpx2`, both transitive — import from `mcp.types` or declare them.
- No cross-clinic test. The isolation looks structurally right, but bar item one is the thing we do not merge without: a clinic-A token asking for a clinic-B `patient_id` must be refused, and it should be pinned. The module is `removable=True`, so an uninstall round-trip against the no-op branch belongs there too.
- `authenticate_token` is described as a shared helper but nothing else uses it — `public.py` still carries its own copy — and the MCP path loses two things that copy does: `last_used_at`, so an admin sees an actively used token as never used and cannot spot a leaked one, and the per-token rate limit, so `initialize`, `tools/list` and bad-token floods are unbounded. Please move both into the helper and have `public.py` call it.
- `docs/adr/0033-mcp-bridge.md` collides with the Poland ADR that merged on Sunday. Renumber to the next free number and set `Status: proposed` — `docs/adr/**` is Ramón's path and an ADR should not be accepted in the PR that proposes it.
- The manifest declares `category: "official"` and `author: "DentalPin Core Team"`. A contributor module should be `community` under your own name unless he adopts it; it changes the support expectation.

It currently conflicts with `main` in `.gitignore`, `alembic.ini`, `pyproject.toml`, `uv.lock` and `docs/modules-catalog.md`, so a rebase is needed once the above is in. I approved the CI run, which had been waiting since Monday.

BabuBahir:
Hi @ZoliQua  & @martinezsalmeron   I fixed the errors and also remove the .gitignore , Let the User read the docs a bit to add the skill to their agent 

Below is picture how a working MCP looks like consuming the 3 tools  , created for this demo.

So below we have claude calling search_patients to get all patients 
 
image

ZoliQua:
Thanks @BabuBahir — the screenshot helps, and a good share of the list is done. Here is exactly where it stands against my last review, checked on `d8110c43`.

**Done ✅**

- The stray `backend/D:/…/modules.json`, `.kilo/`, the `.vs` rules and the `.gitignore` change are all gone from the diff.
- `authenticate_token` is now genuinely shared: `integrations/public.py` calls it, and it stamps `last_used_at` and enforces the per-token rate limit, so the MCP path gets both. That was the most important mechanical point.
- The ADR is renumbered to `0037` with `Status: proposed`.
- The branch no longer conflicts with `main`.

**One red check is mine, not yours.** `frontend-test` failed on a flaky test I added in #479 (`useListQuery … expected undefined to be 'ana'`). Your PR does not touch the frontend; CI runs the merge ref, which pulled my test in. The fix is #483. Sorry for the noise.

**Still open — these are the other three red checks, plus review points:**

1. **`manifest-consistency` / `backend-test`:** `test_every_module_has_valid_manifest` fails because the `mcp` manifest has `role_permissions={}`. Every module must grant at least `admin`, even one with no permissions of its own. `"role_permissions": {"admin": ["*"]}` is enough.
2. **`backend-test`:** `test_rate_limit_exceeded_returns_429` reads `third.json()["detail"]`. The 429 now comes out of the shared helper, and our error envelope puts the text in `message`, not `detail`. Either the test or the response shape needs to follow — whichever matches what the other public endpoints return.
3. **`catalog-freshness`:** run `python backend/scripts/generate_catalogs.py` and commit `docs/modules-catalog.md`.
4. **`uv.lock` is still written by an older `uv`.** `python-lock-freshness` passes, but the file still lacks `revision = 3` and carries 657 `upload-time` fields against 2,207 on `main`. That is why it shows 3,413 changed lines for eleven new packages. Re-running `uv lock` with the repo's `uv` should shrink it to just the additions.
5. **`mcp>=1.0` is still the pin** while the code uses the 2.x lowlevel API; please use `>=2.2,<3`. And `server.py` still imports `mcp_types`, and the test still imports `httpx2`, both transitive — import from `mcp.types` and plain `httpx`, or declare them.
6. **Still no cross-clinic test.** This is the one I will not approve without: a clinic-A token asking for a clinic-B `patient_id` must be refused, pinned by a test. Likewise the uninstall round-trip for a `removable=True` module with its own (no-op) branch.
7. **`category: "official"` / `author: "DentalPin Core Team"`** — still in the manifest. A contributor module ships as `community` under your name unless @martinezsalmeron adopts it.

**Waiting on @martinezsalmeron, not on you:** `server.py:145` still runs every call as `AgentMode.AUTONOMOUS`. Whether an external AI host holding a `patients:write` token may create records with no human in the loop is his decision. Nothing to change until he rules.

When 1–7 are in, mark it ready for review and I will do a full pass.

Issue #101, opened 2026-07-03
[Audit S5 follow-up] Silent-failure sweep: useApi 4xx contract, false-empty reads, save-on-failure text loss, destructive confirms


martinezsalmeron:
Follow-up to #95. Tranche 1 shipped in #100 (useApi \`query\` support fixing the Veri*Factu queue tabs; odontogram error state instead of a fabricated healthy mouth). Remaining S5 items, each needing per-caller review:

## useApi error contract
- 400/404/409/422 currently rethrow with **no toast**; many callers don't catch → silent failure. Define a toast-or-rethrow contract that surfaces the backend \`detail\` on otherwise-unhandled 4xx **without** double-toasting the callers that already handle it (e.g. an opt-out flag, or default-toast + \`suppressErrorToast\`). Audit ~255 call sites.
- Adopt the shared \`errorMessage()\` helper everywhere (many callers read \`data?.message\` while FastAPI sends \`detail\`).

## False-empty reads (render an error state, not a plausible-empty one)
- Agenda: failed fetch renders a blank calendar = looks like a free week (\`useAppointments\`).
- Reports dashboards render 0 € / empty on fetch failure (\`useReports\` pattern).
- Catalog list error → \"no hay elementos\"; budget signature error → \"not signed\".

## Silent data-loss on failed save (preserve input until the mutation resolves)
- Clinical-note composer: parents reset/tear down on a null \`createNote\` result → typed text lost; \`NoteComposer\` also clears \`attachedDocs\` before the save confirms.
- Odontogram treatment-edit modal: \`selectedSurfaces\` edits dropped with a success toast.
- Perio measurements lost inside a closed exam (\`flushPending\` swallows failures).

## Destructive actions without confirmation (no undo endpoint exists)
- Odontogram treatment delete, appointment cancel from modal footer, clinical-note delete (3 of 5 surfaces one-click), recall cancel/mark-done, plan-item delete.

Caveat: module-layer \`.vue\`/\`.ts\` aren't covered by CI lint/typecheck (empty \`modules.json\`), so these need manual dogfooding. See \`docs/technical/audit-2026-07-03.md\` §S5.

PR #462, opened 2026-09-14
feat(staff_attendance): clock in/out events, state, and daily report

lamanji:
## Summary

- 1 commit: `feat(staff_attendance): clock in/out events, state, and daily report`
- **Scope:** module `staff_attendance` | areas `backend, frontend, i18n, infra, testing`.
- **Why:** staff clock in/out events, current state, and a daily pairing report, as a removable module.

## Tasks done

- `feat(staff_attendance)` - clock in/out events (`POST /events`, 409 on consecutive same-kind), state (`GET /status/{user_id}`), daily pairing report (`GET /report`, open shifts flagged), clinic-scoped member picker (`GET /members`), attendance page (`/attendance`, nav order 93), own Alembic branch (`satt_0001`), 10 layer locales, roundtrip uninstall test.

## Modules touched

- `staff_attendance` (new module)

## Checklist

- [x] backend: `ruff check . && ruff format --check .` (backend-lint green)
- [x] frontend: `npm run lint` (frontend-lint green)
- [x] `scripts/dev-verify.sh --fast --with-pytest --module staff_attendance` green (9/10; only local MSYS modules-json-freshness toolchain flake, proven green on CI by PR #461)

## Test plan

- `scripts/dev-verify.sh --with-pytest --module staff_attendance --fast` (compose `db` up)
- Verify end-to-end in the running stack (docker compose up) and confirm the clock flow on `/attendance`.

## Red-conditions

- [x] `backend/tests/modules/staff_attendance/test_staff_attendance.py::test_clock_in_out_and_double_punch_is_409`: red without the consecutive-same-kind guard in `AttendanceService.clock` because a double punch would answer 200 instead of 409.
- [x] `backend/tests/modules/staff_attendance/test_staff_attendance.py::test_non_member_clock_is_404`: red without `AttendanceService._assert_member` because an id-only clock would punch a non-member (and probe user ids across clinics).
- [x] `backend/tests/modules/staff_attendance/test_staff_attendance.py::test_isolation_between_clinics`: red without `clinic_id` scoping in `list_members`/`daily_report` because the other clinic's event and member would leak.
- [x] `backend/tests/modules/staff_attendance/test_uninstall_roundtrip.py::test_staff_attendance_uninstall_roundtrip_is_branch_scoped`: red without the module's own `(staff_attendance,)` Alembic branch because dropping the tables would thread into another module's revision chain.

## Review


ZoliQua (changes requested):
Thanks — this is a clean vertical slice, and the tenancy work is the best I have reviewed in a while: `_assert_member` following payroll's 404-rather-than-403 pattern, a cross-clinic test that covers both the member list and the report, and a per-route 403 test for a role without the grant. That last one is exactly what I keep asking other PRs for. Four things before I can merge.

1. **The clinic timezone is ignored.** `daily_report` and `list_events` bucket on UTC days — `datetime.combine(day, time.min, tzinfo=UTC)`. `Clinic.timezone` is core and documented as the single source of truth for any module needing local-time semantics, and `agenda/tz.py` is the house contract for it (`safe_zone`, `as_utc`), with `agenda/router.py` resolving "today" that way. For a Madrid clinic this report currently runs 02:00 to 02:00 local, and in summer the drift is two hours the other way. Please compute the window from local midnight. If you would rather not depend on `agenda`, `schedules/services/clinic_hours.py` is the low-coupling precedent.

2. **Nothing records who made the punch.** `write` is granted to receptionist, assistant and dentist, and `user_id` comes from the request body, so anyone with the grant can create a punch for any colleague at any timestamp — and `AttendanceEvent` has no `created_by`, the module publishes no events, so `activity_journal` never sees it either. For a working-time record that is the one column that must not be missing. Please store the acting user, and consider publishing `staff_attendance.clocked` with `db=db` so the journal picks it up. Splitting the grant into self-clock and clock-others would be a reasonable follow-up, not required here.

3. **Open and overnight shifts are counted wrong.** A trailing `in` adds `now - open_since` with no cap, so a punch nobody closed keeps accruing hours for days. And `open_since` is only seeded from an `in` inside the queried window, so a shift from 22:00 to 06:00 the next morning reports a runaway open segment on the first day and nothing at all on the second — the closing `out` falls into the `elif slot["open_since"] is not None` branch and is dropped. Please carry the last `in` from before the window, cap the open segment, and add a test for a shift crossing local midnight.

4. **Backdating slips past the 409.** The duplicate guard compares only against the newest event by `at`, so with `in 10:00` and `out 11:00` already recorded, a punch of `{kind: "in", at: "10:30"}` is accepted and the sequence becomes in, in, out. Also `at` is taken as-is: a naive datetime is neither rejected nor interpreted, while the house rule in `agenda/tz.py` is that naive means clinic-local wall clock. Please normalise it and check the neighbours of the inserted time rather than the tail.

Smaller: `clockTarget = ref('')` bound to a `USelectMenu` is the empty-string bite from the contributing notes, please use `null`; `punch()` has no error handling, so the 409 you designed is invisible to the user; `formatHours` hardcodes `h` and `m` with no locale key, which reads oddly under `ar`; and `today` comes from browser UTC rather than the clinic's day.

One thing above this PR that I would like @martinezsalmeron's view on before it goes in: attendance is employee personal data, and our GDPR module is patient-only — its service imports the patient model directly and `gdpr_requests.patient_id` is an FK to `patients.id`, with no plugin hook for another module to register export or erasure. This module has no delete route and no retention rule, so those records sit outside both. I do not think fixing that belongs in this PR, but I would like an issue filed and linked so it is not lost. He needs to review `alembic.ini` and `permissions.ts` here anyway.

ZoliQua (changes requested):
This is a strong pass — all four blocking points are genuinely fixed, and fixed at the right layer.

Verified on `077cc566`:

1. **Timezone.** `_clinic_zone` reads `Clinic.timezone`, `_day_bounds_utc` builds the window from local midnight, and `_as_utc` implements the house rule (naive = clinic wall clock) without importing `agenda`. `test_report_uses_clinic_timezone_not_utc` pins the 00:30 Madrid case that UTC bucketing used to drop. Exactly what I asked for.
2. **`created_by` + journal.** Column, migration `satt_0002_created_by`, the route passes `ctx.user_id`, and `staff_attendance.clocked` publishes with `db=db` so the journal picks it up. `test_clock_stores_created_by` covers it.
3. **Backdating.** The guard now checks both neighbours of the insertion point instead of the tail, and `at` is normalised first. I worked through the sequences by hand: with `in 10:00` / `out 11:00`, a backdated `in 10:30` hits the predecessor and a backdated `out 10:30` hits the successor — both 409.
4. **Smaller items.** `clockTarget` is `undefined`, `punch()` catches and surfaces, `formatHours` goes through `Intl.NumberFormat` units, and `clinicToday()` resolves the day in the clinic's zone. All four done.

Locally: 13 passed in `tests/modules/staff_attendance/` including the uninstall round-trip across both revisions. CI green.

Two things left, one of which is the other half of point 3.

### 1. The overnight fix counts the same shift on both days, and the opening day never closes

The closing day is right now, but the opening day still counts its trailing segment, and it keeps counting it forever. One shift, 22:00 → 06:00 next morning, UTC clinic, queried five days later so nothing is live:

```
DAY 4: 7199s = 2.00h open=True
DAY 5: 28800s = 8.00h open=False
SUM  : 10.00h for an 8.00h shift
```

Two separate problems in that output:

* **`open=True` on day 4 is wrong.** That shift closed at 06:00 on day 5. The report says the person is still clocked in, and will say so for the rest of time, because `open_since` is decided from the window alone and never looks forward for the closing punch. In the UI that is a permanent "still in the building" badge on a historical day.
* **The 2h are counted twice.** Day 4 counts 22:00 → end of day, day 5 counts 22:00 → 06:00. Any aggregation over a range (which a payroll export is going to want, and the module CLAUDE.md lists under Later) is 25% over on this shift.

The cleanest fix is to make the segment arithmetic clip to the window on both ends instead of attributing whole shifts to one day: for each in→out pair, count `max(in, start)` → `min(out, end)`. Day 4 gets 2h, day 5 gets 6h, the sum is 8h, and `open` becomes "no closing punch exists at all", which is a forward query rather than an inference from the window. It also drops the need to carry `open_since` specially — the pre-window `in` you already fetch is just a pair whose start is clipped.

If you would rather keep "attribute the shift to the day it ends", that is defensible too, but then day 4 must not count the trailing segment at all, and `open` still has to be decided by looking for a later `out`.

While you are in there: `_day_bounds_utc` ends the window at `time.max`, which is where the `7199` above comes from — one second lost per open segment. A half-open window (`at < local midnight of the next day`) removes the rounding and the `<=` edge case on a punch at exactly midnight.

Please add a test that asserts the two days sum to the shift length, not just that the closing day is 8h.

### 2. Every 409 raises two toasts

`useAttendance.clock()` calls `api.post(...)` with default options, and `useApi` toasts 400/409/422 itself before rethrowing (#101). `punch()` then catches and calls `toast.add` again, so a double punch pops the same message twice on top of the inline `errorMessage`. The contract in `useApi.ts` is explicit about this case: a caller that surfaces the error itself passes `errorToast: false`. Adding it to the `clock()` post is the one-line fix — please keep the component's toast, since it is the one with the right title.

### Follow-up filed

I opened **#475** for the employee-personal-data gap (`gdpr` is patient-only: three `patient_id` FKs, no contribution hook, while payroll / activity_journal / this module all store employee data). Nothing for you to do in this PR — it is linked so the point is not lost.

@martinezsalmeron still needs to sign off on `alembic.ini` and `frontend/app/config/permissions.ts` here.

ZoliQua:
Cross-reference while you are in there: #474 adds `clinicToday(timeZone)` to `frontend/app/utils/wallClock.ts` (core), doing exactly what your `clinicToday()` in `attendance/index.vue` does — and it reads the Y-M-D fields off the wall-clock `Date` rather than going through `Intl.DateTimeFormat` parts. Whichever of the two lands first, the second should import the core helper instead of keeping its own copy. Same story on the backend: your `_clinic_zone` is now the third copy of `Clinic.timezone` resolution (`agenda/tz.py`, #474's `payment_gateways/service.py`, yours) — I have asked @martinezsalmeron whether that belongs in `app/core/`, since `Clinic` is a core model. Nothing to change here until he answers; just flagging so you do not find out by collision.

lamanji:
@ZoliQua round 3 addressed in `08c4d91f` (CI 12/12 green at head).

1. Overnight double-count + permanent open flag. The report now clips every in→out pair to the window on both ends (`service.py:220`, `max(in, start)` to `min(out, end)` at `:273-277`), `open` is a forward query for any later `out` (`:283`), and the window is half-open (`:52-54`, no more `time.max`). `test_overnight_shift_sums_exactly_across_both_days` (`test_staff_attendance.py:309`) asserts the two days sum to the shift length.
2. Double toast. `clock()` in `useAttendance.ts:40` passes `errorToast: false`; the component toast stays as the single surface.

On the #474 cross-reference: agreed on the direction (import the core `clinicToday` / shared zone helper, whichever lands first). No action from me until the maintainer answers the ownership question.

`alembic.ini` and `permissions.ts` still need maintainer sign-off.

martinezsalmeron:
@lamanji thank you for this one, and for how quickly you turned round 2 around. I checked `08c4d91f` locally (14 passed) and everything @ZoliQua asked for is genuinely done: the clipped pairs sum exactly across both days, `open` is a real forward query, the window is half-open, and the 7-day lookback bound is a nice touch nobody asked for. `alembic.ini` and `permissions.ts` are fine on my side, so consider that sign-off given. The tenancy work and the per-route 403 test are exactly the standard I'd like every module to follow.

While testing in the browser and against the running stack I found three more things. None of them were in the earlier reviews, so they're on me for arriving late, not on you. All three are small and live in `service.py` and `punch()`:

1. **A future `at` can lock a user out.** `clock()` accepts e.g. an `out` dated 2030 (easy typo). After that the user can clock in, but every `out` answers 409 forever because the successor is the 2030 punch, and there's no void route to undo it. Rejecting `at > now + a small skew` with a 422, plus one test, would close it.
2. **A shift opened yesterday and still open is missing from today's report.** `by_user` is seeded only from events inside the window, so someone who clocked in at 22:00 yesterday and hasn't left yet doesn't appear today at all. Seeding also from users whose last pre-window event is an unclosed `in` should do it — the lookback query you already have is most of the way there.
3. **The remaining 409 toast shows the raw error.** Now that `useApi`'s toast is off, the one left reads `FetchError: [POST] "…/events": 409 Conflict`. `punch()` reads `e.data.detail`, but our API envelope returns `message`. `errorDetail()` in `frontend/app/utils/error.ts` handles both, so it's a one-line swap.

Smaller, only if you're in there anyway — happy to take any of these as a follow-up instead:

- The day feed shows the badge and the name but not the punch time, which is the one thing a receptionist will look for.
- `docs/technical/staff_attendance/events.md` still says the module publishes no events, and the module `CLAUDE.md` doesn't list the `staff_attendance.clocked` payload yet.
- Since the module has never shipped, `satt_0002` could be folded into `satt_0001` with `created_by` as NOT NULL. Not a blocker.

Once 1–3 are in I'll merge straight away. Thanks again for the care you're putting into these modules — it shows.

lamanji:
All three are in `195dd46c`, plus two of the three smalls. Thank you for the sign-off on `alembic.ini` and `permissions.ts`, recorded.

1. Future `at`. `clock()` rejects anything past now + 5 min skew with 422 "Punch time is in the future" (`service.py:29,134-139`). `test_future_punch_is_422` (`test_staff_attendance.py:361`) pins it.
2. Yesterday's open shift. The lookback now runs clinic-wide instead of only for in-window users, and a last pre-window punch that is an unclosed `in` seeds the member (`service.py:257-285`). `test_open_shift_from_yesterday_appears_today` (`:377`) asserts one row, open, 12h on the next day.
3. Raw 409 toast. `punch()` reads `errorDetail(e)` (`index.vue:9,99`); the component toast stays the single surface.

Smalls: the day feed shows the punch time in clinic tz (`index.vue:67,183`); `events.md` lists `staff_attendance.clocked` with its consumer and the module CLAUDE.md documents the payload. I left the `satt_0002` fold as a follow-up: the chain is signed off and green, and the fold needs roundtrip-test rework, so it deserves its own commit, not a ride-along.

Verified: module suite 15 passed (13 existing + 2 new) plus the uninstall roundtrip, `dev-verify.sh --fast` 9 passed. One honest caveat: local `nuxt typecheck` in my worktree reports 276 pre-existing tree-wide resolution errors on Windows (0 in this module's files, eslint clean), so the typecheck verdict belongs to CI.

ZoliQua (approved):
Approving — both of my round-2 points are fixed, and I re-ran the measurement that found the first one rather than taking it on trust.

**The overnight shift.** Same fixture as before — UTC clinic, `in` 22:00 on day 4, `out` 06:00 on day 5, queried five days later — against `195dd46c`:

```
DAY 4: 7200s open=False
DAY 5: 21600s open=False
SUM  : 8.00h for an 8.00h shift
```

Was 2h + 8h = 10h with day 4 permanently `open=True`. The clipped pairs sum exactly, `open` is a real forward query, and the half-open window also got rid of the lost second (`7199` → `7200`).

**The double toast.** `clock()` passes `errorToast: false`, the component toast is the single surface, and with Ramón's `errorDetail(e)` point in it reads the envelope's `message` instead of the raw `FetchError`.

Locally on a fresh DB: `tests/modules/staff_attendance/` = 16 passed, uninstall round-trip included. CI green, branch mergeable. Ramón's three round-3 points (future-punch 422, yesterday's open shift seeding today, `errorDetail`) are all in with a test each for the first two; I leave the final word on those to him since they are his. Agreed on leaving the `satt_0002` fold for its own commit.

martinezsalmeron (approved):
Approving and merging — all three round-3 points are in, and I re-ran the probes that found them rather than reading the diff only. Thanks @lamanji for the fast, precise turnaround, and @ZoliQua for re-measuring the overnight case.

Checked on `195dd46c` merged with current `main` (no file overlap with anything that landed since):

1. **Future `at`.** `POST /events` with an `out` dated 2030 → `422 Punch time is in the future`; a punch 3 min ahead (inside the 5 min skew) still goes through. The user can no longer be wedged into permanent 409s.
2. **Yesterday's open shift.** Clocked a member in yesterday evening and left it open: they now appear on today's report as open, and the hours split exactly at the clinic's local midnight (my demo clinic is `America/New_York`: 8.00h yesterday + the rest today, summing to the elapsed time). Making the lookback clinic-wide instead of per-seen-user was the right call.
3. **409 toast.** In the browser a double clock-in shows exactly one toast, reading "Already clocked in" — no raw `FetchError`. Clock-out then closes the shift and the feed shows each punch time in the clinic's zone.

Module suite locally: 15 passed. CI 12/12 green.

Non-blocking, for the follow-up commit you already planned for the `satt_0002` fold:

- The nav entry has no `"section"`, so since #467 "Attendance" lands in the top group right under Agenda instead of under Practice. `"section": "practice"` (same as `staff_tasks` / `reports`) fixes it.
- The 409/422 messages come through in English regardless of UI locale — same as most modules today, so not something to solve here.
- Screen docs in `docs/user-manual/{en,es}/staff_attendance/` could mention the punch time in the feed and get `last_verified_commit` bumped.

#475 (GDPR for employee data) stays open as the tracking issue for export/erasure.

PR #463, opened 2026-09-15
feat(treasury): cash and bank accounts with transfers and corrections

lamanji:
## Summary

- 3 commit(s): `feat(treasury): cash and bank accounts with transfers and corrections`; `Merge remote-tracking branch 'upstream/main' into pr/treasury-module`; `fix(treasury): correct composable import depth (CI frontend-typecheck TS2307)`
- **Scope:** modules `treasury` | areas `backend, frontend, i18n, infra, testing`.
- **Why:** feat(treasury): cash and bank accounts with transfers and corrections

## Tasks done

- `feat(treasury): cash and bank accounts with transfers and corrections` - treasury
- `Merge remote-tracking branch 'upstream/main' into pr/treasury-module` - core
- `fix(treasury): correct composable import depth (CI frontend-typecheck TS2307)` - treasury

## Modules touched

- `treasury`

## Checklist

- [x] `scripts/dev-verify.sh` green (confirm + tick; CI is the authority)

## Test plan

- `scripts/dev-verify.sh --with-pytest --module treasury --fast` (compose `db` up)

## Red-conditions (L76)

Fill before shipping (one line per guard-adjacent test):

- [x] `backend/tests/modules/treasury/test_treasury.py::test_duplicate_name_is_409`: red without the clinic-scoped `name` UNIQUE constraint + `IntegrityError`→409 guard in `TreasuryService.create_account` (L6: no select-then-insert) because a duplicate name would otherwise surface as a 500.
- [x] `backend/tests/modules/treasury/test_treasury.py::test_same_account_transfer_is_422`: red without the `from_account.id == to_account.id` guard in `TreasuryService.transfer` because pairing two legs on the same account would double-count that account's balance.
- [x] `backend/tests/modules/treasury/test_treasury.py::test_foreign_clinic_account_is_invisible`: red without the `clinic_id == clinic_id` filters in `get_account`/`list_accounts`/`statement` (L1) because an id-only query would leak another clinic's ledger rows.
- [x] `backend/tests/modules/treasury/test_uninstall_roundtrip.py::test_treasury_uninstall_roundtrip_is_branch_scoped`: red without the per-module Alembic branch (`treasury`, `tre_0001`, `down_revision="0001"`, registered in `alembic.ini`) because a threaded cross-module branch cannot be removed without disturbing other modules' chains.

## Review


ZoliQua (changes requested):
Thanks — the shape is right: append-only entries, paired legs sharing a `group_id`, balances derived rather than stored (which matches the deliberate call in billing), `Numeric(12,2)` on convention, an own Alembic branch with a round-trip test, and an honest "Later" list. Four things before I can merge.

1. **Money moves with no record of who moved it.** The service comment says "the audit trail lives in activity scope", but `TreasuryEntry` has no `created_by`, the module publishes no events, and `activity_journal` only records event types it subscribes to — so nothing anywhere knows who made a transfer or a correction. A correction captures the amount, the direction and a mandatory memo, and then loses the one field that makes it auditable. Please store the acting user on the entry, and ideally publish `treasury.transferred` and `treasury.corrected` with `db=db` so the journal picks them up.

2. **`DELETE /accounts/{id}` takes the ledger with it.** `db.delete(row)` against a foreign key declared `ondelete="CASCADE"` means one call erases an account's entire financial history, and the code's own comment already says to prefer `is_active=False`. Let us enforce that: refuse with 409 when the account has entries. Deactivation is already supported in `AccountUpdate`.

3. **Unknown kinds vanish silently.** `_SIGN.get(kind, Decimal("0"))` means any value outside the map contributes zero to the balance with no error anywhere, and there is no CHECK constraint on `kind` or on `amount > 0` at the table level — payments has both. Please add the constraints and raise rather than default.

4. **The decimal edges are open.** `amount` and `opening_balance` carry no `max_digits`/`decimal_places`, so `10.005` is accepted and quietly rounded by the column, and an oversized `opening_balance` raises a `DataError` that `create_account` does not catch (it handles only `IntegrityError`), which surfaces as a 500. `Field(gt=0, max_digits=12, decimal_places=2)` plus an explicit quantize in the service closes both.

On tests: #462 ships a per-route 403 for a role without the grant and an HTTP-level cross-clinic assertion; this one has neither — the foreign-clinic test stops at the service layer. With `admin: ["*"]` as the only grant, the refusal path is the whole permission story, so please mirror both.

Frontend: the page has no `definePageMeta({ middleware: ['auth'] })` and no `can()` guard, so a direct `/treasury` URL renders an empty shell of 403s; `transferFrom`, `transferTo` and `selectedId` are `ref('')` on a `USelectMenu`, which is the empty-string bite from the contributing notes — please use `null`; `create()` and `doTransfer()` swallow 409 and 422, so the modal just closes; and balances render as bare strings with no currency, although `Clinic.currency` is core and there is a Babel formatter in `app/core/utils/currency.py`. The user manual also documents a **Correct** action the screen does not offer — either build it or adjust the text.

Two things I would rather @martinezsalmeron ruled on than decide myself. First, `payments` already owns cash-in with a `method` that includes `cash`, and the product already calls that "la caja real de la clínica"; with auto-posting deferred, treasury's balances are a parallel ledger maintained by hand and disconnected from every real movement in the system. That is a product-shape call. Second, the tests assert a cash drawer at −25.50: I accept the declared "no overdraft guards in v1", but for a cash account specifically I would want either a guard or a visible warning. He needs to review `alembic.ini` and `permissions.ts` here in any case.

ZoliQua (changes requested):
All four blocking points are fixed, and the extras I listed are in too. Verified on `1b91304d`:

1. **`created_by` + events.** Column on `TreasuryEntry`, migration `tre_0002_created_by`, routes pass `ctx.user_id`, and `treasury.transferred` / `treasury.corrected` publish with `db=db`. `test_transfer_stores_created_by` covers it.
2. **Delete no longer eats the ledger.** `delete_account` counts entries first and answers 409 with a message that points at deactivation. Covered at both layers (`test_delete_account_with_entries_is_409` and the HTTP variant), plus the empty-account happy path.
3. **Unknown kinds.** `_signed` raises instead of defaulting to zero, and `tre_0003_entry_guards` adds `ck_treasury_entries_kind` and `ck_treasury_entries_amount_positive`, so the map and the table now agree.
4. **Decimal edges.** `max_digits=12, decimal_places=2` on all three money fields plus `quantize` in the service; `test_decimal_edges_are_422_not_500` pins the 500 that used to happen.

Tests: `test_denies_roles_without_treasury_grant` and `test_cross_clinic_accounts_are_invisible` are both there now, at HTTP level. Frontend: `definePageMeta({ middleware: ['auth'] })`, the `can()` guard, `undefined` on the selects, `Intl.NumberFormat` with `currentClinic.currency`, and the Correct modal the manual promised. Locally 13 passed in `tests/modules/treasury/` including the round-trip across all three revisions.

Two things left, both small.

### 1. Every 400/409/422 raises two toasts

`useApi` toasts those three statuses itself and then rethrows (`useApi.ts:135-143`, and the comment block at :22-29 spells out the contract). The three POSTs in `useTreasury.ts` (:31, :39, :56) take the default, and `fail()` in `index.vue` toasts again on the way out — so a duplicate account name pops the same message twice.

The documented fix is `errorToast: false` on those three calls, keeping `fail()` as the single surface, since it is the one with the right title. Please do not remove `fail()` instead: dropping the component toast *and* passing `errorToast: false` is how you get a silent 409, which is a mistake I made myself in #457 and Ramón had to fix on main.

Same pattern in #462, same one-line fix.

### 2. `GET /accounts` loads the whole ledger, once per account

`_with_balance` calls `TreasuryService.balance` per row, and `balance` selects every entry for that account and sums in Python. So the first render of the page is `1 + N` queries, and query `i` returns every movement that account has ever had. A clinic cash drawer with a handful of movements a day passes a million rows in a few years, and this is the page's default load, not a report.

The balance is expressible directly:

```sql
SELECT account_id, SUM(CASE kind WHEN 'transfer_out' THEN -amount ... END)
FROM treasury_entries WHERE clinic_id = :c GROUP BY account_id
```

one query for the whole list, with the sign map living next to `_SIGN` so the two cannot drift. Worth doing before it has data in it rather than after.

Smaller, not blocking: `transfer` does not check that either account is `is_active`, so money can be moved into an account that was deactivated precisely to stop that; and `at` accepts a naive datetime and stores it as UTC, whereas #462 now treats naive as clinic wall-clock — two modules by the same author answering the same question differently is the kind of thing that is cheap to align now.

@martinezsalmeron — my two questions from the first review are still open and they are yours, not lamanji's: whether a hand-maintained ledger parallel to `payments` (which already owns cash-in) is the product shape we want with auto-posting deferred, and whether a cash account should guard or at least warn on a negative balance. `alembic.ini` and `permissions.ts` need your sign-off here as well.

lamanji:
@ZoliQua round 3 addressed in `bd8ff754` (CI 12/12 green at head).

1. Double toasts. All three POSTs in `useTreasury.ts:35,44,62` pass `errorToast: false`; `fail()` in `index.vue:54` stays the single surface.
2. N+1 balances. `GET /accounts` now aggregates in one query (`service.py:88-100`, `GROUP BY` plus signed `SUM`); the sign map is derived from `_SIGN` (`:30`) so SQL and Python cannot drift.
3. Smaller items. `transfer` rejects when either account is inactive (`service.py:184,242`); naive `at` is treated as clinic wall-clock (`:55`), aligned with #462.

The two product-shape questions (ledger parallel to payments, negative-balance guard) plus `alembic.ini` / `permissions.ts` remain open for the maintainer.

martinezsalmeron:
@lamanji thank you so much for this one, and for the patience across three rounds. 🙏 And @ZoliQua, thank you for a genuinely excellent review — it made mine easy.

I reviewed `bd8ff754` end to end: read the whole diff, installed the module locally and probed every route against the running stack.

## What is great

The backend is solid, and I mean that. Everything behaved exactly as it should:

- duplicate name → 409; `10.005` and `25,50` → 422; same-account or inactive-account transfer → 422; correction without memo → 422; delete with entries → 409
- balances are right and come from a single aggregate query
- with `activity_journal` installed, both `treasury.transferred` and `treasury.corrected` land in the journal **with the actor** — the audit story works
- install → uninstall round-trips cleanly (tables gone, other branches untouched)

Architecture is respected throughout: no cross-module deps, FKs to core only, own Alembic branch, transactional events (`db=db`), namespaced admin-only permissions, `auto_install=False`. All of ZoliQua's blocking points from rounds 1–3 are truly fixed.

`alembic.ini` and `permissions.ts`: both fine — signed off. ✅

## Three small things before merge

We are really close. These are all quick:

1. **The error toast shows a raw `FetchError`.** `fail()` in `pages/treasury/index.vue:54` reads `e.data.detail`, but our API puts the text in `message` (`{"data":null,"message":"An account with this name already exists",...}`). Since round 3 (correctly!) made `fail()` the only toast, a duplicate name now shows `FetchError: [POST] … 409 Conflict`. The fix is one line: use `errorDetail()` from `frontend/app/utils/error.ts`, which already handles `detail`/`message` and the 422 list case. Totally understandable miss — it is the same one as in #462, and the helper is easy to overlook.
2. **Nav section.** Please add `"section": "financials"` to the navigation entry — `docs/technical/creating-modules.md` asks every new module to self-place (#232).
3. **Docs caught up with the code.** `docs/technical/treasury/events.md` still says the module publishes no events (it publishes two now 🎉); the module `CLAUDE.md` still says `DELETE` cascades entries (it is a 409 now) and should document the two event payloads.

## Suggestions (not blocking — happy to take them here or in a follow-up)

- The statement shows no date and no sign, so a 25.50 out looks identical to a 25.50 in. A date column and a `−`/colour on outgoing legs would make it read like a real ledger.
- A Spanish user will type `25,50` and get a 422. Normalising `,` → `.` before sending would be kind to them.
- Inactive accounts still appear in the transfer selects and then 422 — filtering them out avoids the dead end.
- There is no UI for the opening balance or for deactivating an account (the API supports both).
- Tiny ones: the `opening` kind is never written; `tre_0002`/`tre_0003` could fold into `tre_0001` since the module was never deployed; `POST /transfers` returns 200 rather than 201; a future `at` is accepted.

## The two product questions ZoliQua left for me

- **A ledger parallel to `payments`:** I am fine with it. The module is optional, community, manually installed, and clearly scoped as "where the money sits". Auto-posting from payments/expenses deserves its own design discussion, exactly as your "Later" list says — good call keeping it out of v1.
- **Negative cash balance:** a visible warning rather than a guard, please — e.g. the balance in red when it is below zero. Blocking a drawer that is simply mis-counted would get in the way more than it helps. Non-blocking for this PR, but very welcome.

Once 1–3 are in I will merge straight away. Thanks again to you both — this is a nice, clean module and the collaboration on it has been a pleasure to read. 🚀

lamanji:
All three blockers are in `7f721882` (plus stamp `ba6764f0`), with several suggestions taken. Thank you for the sign-offs and both product rulings, all recorded.

1. Raw toast. `fail()` reads `errorDetail(e)` (`index.vue:10,57`); it stays the single surface.
2. Nav section. `"section": "financials"` on the entry (`__init__.py`, same as billing/budget); backend-manifest gate green.
3. Docs. `events.md` lists both published events with their consumer; the module CLAUDE.md documents both payloads and now states DELETE is 409. `ba6764f0` stamps the verified SHAs on both screen docs.

Suggestions taken: POST /transfers answers 201 (`router.py`, tests `:113,:194` updated); future-dated transfers/corrections are 422 (`service.py:32,198,261`, `test_future_movement_is_422` at `:403`); transfer pickers list active accounts only (`index.vue:45,292,299`); amounts accept `25,50` (`:62,109,119`); statement rows show dates with signed red out-legs (`:66,81,235-239`); negative balances render red, never block (`:205`).

Left as follow-up: opening-balance and deactivate UI (new flows, not smalls), and the tre_0002/tre_0003 fold (the chain is signed off and green; it needs roundtrip-test rework, so its own commit). The `opening` kind stays in `_SIGN` unwritten on purpose: opening lives on the account row by design, and the map must keep vocabulary parity with the check constraint.

Verified: module suite 16 passed plus the uninstall roundtrip, `dev-verify.sh --fast` 9 passed, eslint and ruff clean. Same honest caveat as #462: local `nuxt typecheck` in my worktree reports pre-existing tree-wide resolution errors (0 in treasury files), so the typecheck verdict belongs to CI.

ZoliQua (approved):
Approving — both of my round-3 points are fixed, and fixed the way I suggested.

**Balances.** `TreasuryService.balances` is one `SUM(amount * CASE …) GROUP BY account_id` for the whole clinic, and the negative kinds are derived from `_SIGN` rather than restated in SQL, so the map and the query cannot drift. `GET /accounts` passes that dict into `_with_balance` — no per-account query, no entry rows loaded into Python.

**Single toast.** All three POSTs in `useTreasury.ts` pass `errorToast: false`, `fail()` stays the one surface, and with Ramón's point it now reads `errorDetail(e)` so a duplicate name shows the message instead of the raw `FetchError`.

Locally on a fresh DB: `tests/modules/treasury/` = 17 passed, uninstall round-trip included. CI green, branch mergeable. Ramón's three blockers are his to close; from my side nothing is left. Nice work across four rounds.

martinezsalmeron:
Merged — thank you @lamanji, and thank you @ZoliQua for the review rounds. 🚀

I re-reviewed `7f721882` + `ba6764f0` and tested in the browser on top of current `main`: the duplicate-name 409 now shows one readable toast, Treasury sits under Financials, `25,50` is accepted, the statement shows dates with signed red out-legs, a negative balance is red without blocking, and both `treasury.transferred` / `treasury.corrected` land in the activity journal with the actor. Docs match the code.

Two commits of mine landed on your branch, so you know what they are:

- `50dcc4b8` — merge of `main`. #462 went in first and conflicted on the registry files (`alembic.ini`, `pyproject.toml` entry points, `activity_journal` subscriptions, `permissions.ts`, catalogs). All additive, both sides kept, catalogs regenerated. Treasury + staff_attendance + activity_journal suites (41) and the uninstall round-trips pass on the merge.
- `2b9793ed` — unrelated to treasury: `useListQuery.test.ts` slept a fixed 400 ms against a 300 ms debounce and was flaking `frontend-test` on `main` too. It now polls with `vi.waitFor`.

Small follow-ups, whenever you like (none blocking):

- The Correct modal labels the memo "(optional)", but it is required there — with an empty memo the button silently does nothing. A dedicated label (or a disabled button) would fix it.
- The header's three buttons have no `flex-wrap`, so they will overflow on a phone.
- The transfer selects have no placeholder, so they render as two tiny empty boxes.
- `normAmount` swaps only the first comma, so `1.234,50` still 422s.
- Already agreed: opening-balance / deactivate UI, and folding `tre_0002`/`tre_0003`.

PR #464, opened 2026-09-15
feat(agenda): QR check-in with signed short-lived tokens and public page

lamanji:
## Summary

- 3 commit(s): `feat(agenda): QR check-in with signed short-lived tokens and public page`; `Merge remote-tracking branch 'upstream/main' into pr/agenda-qr-checkin`; `fix(agenda): session-cookie auth for check-in QR download and publicRoutes indent`
- **Scope:** modules `agenda` | areas `backend, frontend, i18n, testing`.
- **Why:** feat(agenda): QR check-in with signed short-lived tokens and public page

## Tasks done

- `feat(agenda): QR check-in with signed short-lived tokens and public page` - agenda
- `Merge remote-tracking branch 'upstream/main' into pr/agenda-qr-checkin` - core
- `fix(agenda): session-cookie auth for check-in QR download and publicRoutes indent` - agenda

## Modules touched

- `agenda`

## Checklist

- [ ] module `CHANGELOG.md` + docs updated per CLAUDE.md checklists
- [ ] catalogs + `frontend/modules.json` regenerated

## Test plan

- `scripts/dev-verify.sh --with-pytest --module agenda --fast` (compose `db` up)

## Red-conditions (L76)

Fill before shipping (one line per guard-adjacent test):

- [x] `backend/tests/test_appointment_checkin.py::test_checkin_unknown_appointment_is_404`: red without the token/appointment resolution in the public endpoint because a raw minted token for a nonexistent appointment would otherwise 500 or 200 (`AppointmentService.get_appointment` would return None unguarded)
- [x] `backend/tests/test_appointment_checkin.py::test_checkin_wrong_state_is_422`: red without the `checked_in`-allowed state guard because any appointment, terminal included, would otherwise accept the transition through the canonical machine
- [x] `backend/tests/test_appointment_checkin.py::test_checkin_token_is_tenant_bound`: red without the clinic-cross-check because the token's `clinic_id` must gate the lookup to keep the multi-tenant boundary even on the public (unauthenticated) path
- [x] `backend/tests/test_appointment_checkin.py::test_public_checkin_is_rate_limited`: red without the public-endpoint rate limit because an unauthenticated POST path with no login/lease would otherwise be an open brute-force surface

## Review


ZoliQua (changes requested):
Thanks — the shape is right, and the tenant binding is the part I look at first on a public route: the `clinic_id` travels in the token, every lookup goes through `get_appointment(db, clinic_id, appointment_id)`, and there is a test for the cross-clinic case. The response is also properly lean, just `appointment_id` and `status`, so a stranger holding a link learns nothing about the patient. Four things before I can merge.

1. **Please move the public page into the agenda layer.** Both existing patient-facing pages live in their own module under the already-public `/p/` prefix: `notifications/frontend/pages/p/push/[token].vue` and `budget/frontend/pages/p/budget/[token].vue`, both with `layout: 'public'`. Neither had to touch `frontend/app/middleware/auth.global.ts`. As written this adds a core-owned `frontend/app/pages/check-in.vue`, widens `publicRoutes` with `/check-in`, and picks `layout: 'guest'`, which is the staff login chrome rather than the patient one. Following the precedent — `backend/app/modules/agenda/frontend/pages/p/check-in/[token].vue` — reverts the middleware change, keeps agenda uninstallable without leaving a dead route and ten orphan locale blocks in core, and moves the token out of the query string into the path, where proxy access logs and `Referer` cannot pick it up.

2. **Do not sign a public token with `SECRET_KEY`.** `checkin.py` uses `settings.SECRET_KEY`, the same key that mints `dp_access`. We already decided this the other way for budget: `BUDGET_PUBLIC_SECRET_KEY` with a dev fallback and a production warning at boot. A token that gets printed on paper and scanned by strangers should not share a key with the session tokens. Please add a dedicated secret with the same pattern, and put it in `.env.example`.

3. **`test_public_checkin_is_rate_limited` cannot fail.** `limiter` is built as `Limiter(..., enabled=_limiter_enabled)` with `_limiter_enabled = ENVIRONMENT == "production" and not TESTING`, so it is off under pytest, and the test's only assertion is `assert 401 in statuses` — it passes with the decorator deleted. Either force-enable the limiter for that one test and assert a 429, or drop it from the red-conditions list so it does not read as a guard. Worth matching budget and webpush while you are there: both add a per-token `key_func` next to the per-IP limit, which matters once the token is a path segment.

4. **`origin` on the QR endpoint accepts any host.** Only the scheme is checked, so a user with `agenda.appointments.write` can mint a QR pointing at `https://evil.example/check-in?t=<valid token>` and hand it to a patient. The endpoint does not redirect, but the phone that scans the image does. Please build the URL from `settings.allowed_origins_list` server-side and drop the parameter.

Smaller: pass `note="qr-checkin"` to `transition()` so the status event is attributable (it is `changed_by=NULL` today); move the resolve-transition-commit sequence out of the router into a service method, as budget and webpush both do; and the public POST eager-loads the patient, professional and the whole treatments graph for a response that returns two fields.

One thing worth stating in the description rather than leaving implicit: the webpush token burns atomically (`UPDATE … WHERE used_at IS NULL RETURNING`), while this one stays replayable for its full fifteen minutes. I think that is defensible for an idempotent `checked_in`, but let us say so.

Docs nit: the screen files list `check-in-token` under `related_endpoints` but not `check-in-qr` or `public/check-in`, and `last_verified_commit` was not bumped.

`frontend/app/**` is a CODEOWNERS path, so this needs @martinezsalmeron too — though if point 1 lands, those two files drop out of the diff entirely.

ZoliQua (changes requested):
Good round — three of the four are fully done, and the smaller items too. First, a correction I owe you.

**I was wrong about the middleware.** I wrote that `/p/budget` and `/p/push` "neither had to touch `frontend/app/middleware/auth.global.ts`". They both did: `91f23360` added `'/p/budget'` to `publicRoutes` and `4380aec6` added `'/p/push'`. Adding `'/p/check-in'` is the precedent, not a deviation — keep it. Sorry for sending you looking.

Verified on `d7de662f`:

2. **Dedicated secret.** `AGENDA_PUBLIC_SECRET_KEY` in `config.py` and `.env.example`, with `_checkin_secret()` falling back to `SECRET_KEY` in dev and warning once in production — the same shape as budget, comment included.
3. **The rate-limit test can fail now.** `monkeypatch.setattr(limiter, "enabled", True)` and the assertion is `429 in statuses`. You also added the per-token `key_func` (`5/15minute`) next to the per-IP limit, which is the part that matters most now the token is a path segment.
4. **The QR origin is server-side.** The parameter is gone; the URL is built from `settings.allowed_origins_list[0]` with an explicit 422 when `ALLOWED_ORIGINS` is unset or not an http(s) origin. A caller can no longer point a printed code at a host they chose.

Smaller items, all in: `note="qr-checkin"` on the transition, `AppointmentService.public_checkin` holds resolve → transition → commit so the router is thin, the public read is lean (plain `select(Appointment)`, no patient/professional graph), the screen doc lists all three endpoints with a real `last_verified_commit`, and the replay window is now stated explicitly in the route docstring rather than left implicit. Locally `tests/test_appointment_checkin.py` = 10 passed. CI green.

Two things left.

### 1. The old core page is still in the diff

`backend/app/modules/agenda/frontend/pages/p/check-in/[token].vue` is the right page — `layout: 'public'`, token in the path, agenda-owned. But `frontend/app/pages/check-in.vue` is still there too: the original `layout: 'guest'` version that reads the token from `route.query.t` and POSTs it in a body.

It is now unreachable — `publicRoutes` has `/p/check-in` and no longer `/check-in`, so the global middleware sends it to `/login` — which makes it dead code rather than a live leak. Still, it is a duplicate of a security-sensitive page sitting in a CODEOWNERS path, and the next person to read it will not know which one is real. Please delete it. That also drops `frontend/app/**` from the diff and removes a reviewer from your critical path.

(The ten `appointments.checkin` locale blocks in core are fine as they are — agenda has no layer `i18n/` directory at all, all 84 `appointments.*` keys already live in core, so you are following the module's existing pattern, not creating an orphan. My original wording lumped these in with the page; only the page needs to go.)

### 2. `AGENDA_PUBLIC_SECRET_KEY` is not wired into the production compose

`docker-compose.prod.yml` passes `BUDGET_PUBLIC_SECRET_KEY` through, so a deploy can actually set it. The new key has no such line, which means a production deploy on that compose file cannot set it at all and always lands on the `SECRET_KEY` fallback plus the warning — exactly the state the key exists to prevent. One line next to the budget one.

Everything else is ready as far as I am concerned; with the core page gone this only needs @martinezsalmeron for `backend/app/config.py`.

lamanji:
@ZoliQua round 3 addressed in `6e7735ad` (CI 12/12 green at head).

1. Dead core page gone. `frontend/app/pages/check-in.vue` deleted (62 lines); the agenda-owned `p/check-in/[token].vue` is the only page. The remaining `frontend/app/**` touch is just the `/p/check-in` line in `publicRoutes` (`auth.global.ts:34`), kept per your correction. Core locale blocks untouched.
2. Prod compose. `AGENDA_PUBLIC_SECRET_KEY` wired in `docker-compose.prod.yml:36`, next to the budget line.

`backend/app/config.py` still needs maintainer sign-off.

martinezsalmeron:
Thanks a lot @lamanji — and thanks @ZoliQua for three very thorough rounds. This one has come a long way, and the security posture is genuinely good now. I went through it end to end on `6e7735ad`, running stack included.

**What I verified live, all working nicely:**

- Mint and QR sit behind `agenda.appointments.write` (401 without a session); the stale `?origin=` parameter is ignored, as intended.
- `/p/check-in/<token>` in the browser: loading → "check-in completed", `confirmed → checked_in`, and the status event carries `note="qr-checkin"`.
- Re-scanning the same token is a calm 200; garbage tokens are 401; a `dp_access` session token presented as a check-in token is 401 too.
- Tenant binding through the token's `clinic_id`, lean public read, thin router. `AGENDA_PUBLIC_SECRET_KEY` in `backend/app/config.py` mirrors the budget key — that has my sign-off.

**One thing blocks the merge, and it is a one-liner:**

`CheckinQrDialog.vue:31` still builds the copyable link as `${window.location.origin}/check-in?t=${token}`. That page was (rightly) deleted in round 3, so a patient who receives the copied link lands on the staff login screen — I confirmed it by opening that URL. It should be `/p/check-in/${token}`. Even better, if you feel like it: let the mint endpoint return the `url` built server-side with the same origin logic as the QR, so the link and the QR can never point at different hosts (`window.location.origin` vs `ALLOWED_ORIGINS[0]`) and the frontend builds no URLs at all.

**Small leftovers from round 2, while you are in there:**

- `useAppointments.ts:208` still sends `?origin=…` to the QR endpoint, and the comment above it describes the old behaviour.
- `render_checkin_qr_base64` in `checkin.py` has no callers.
- `docs/technical/agenda/permissions.md` does not list the three new endpoints, and the module `CLAUDE.md` does not mention check-in yet.
- The `fix(review): …` CHANGELOG entry describes the review process rather than the product change — folding it into the `feat:` entry reads better for whoever scans the release notes.
- Not for this PR, just noting it: there is no date guard, so I could check in an appointment dated 21 Sep today. Low risk since staff mints the token and it lives 15 minutes; a follow-up is fine.

**A product observation, which matters more than any of the above:**

As built, reception has to open each appointment, press the QR button and show the code to the patient — that is more clicks than simply pressing "check-in" themselves, so I am not sure a clinic would reach for it day to day. Where this feature really earns its place is in a **kiosk-style flow**: a fixed QR (or a tablet) at the entrance that patients use on their own, with nobody at the desk involved. The 15-minute, staff-minted, per-appointment token cannot serve that case.

The good news is that almost everything here is the right foundation for it — the signed-token helper, the public page, the rate limiting, `public_checkin` going through the canonical status machine. The hard part of kiosk mode is a different one: how a patient identifies themselves on an unauthenticated surface without leaking who has an appointment today. That deserves a short design discussion before any code, so I would suggest we open an issue for it rather than stretch this PR.

So: fix the link, sweep the leftovers, and let us talk about the kiosk shape in an issue. Really appreciate the patience through the rounds — the result shows it.

martinezsalmeron:
Opened the design issue for the kiosk flow: #480. Ideas very welcome there — no rush, and no need to fold any of it into this PR.

lamanji:
The link now comes from the server in `1ec47ab9` - I took the "even better" option.

The origin logic lives in one helper, `_checkin_base_url()` (`router.py:603`), used by both the mint endpoint and the QR render. Mint returns `url` alongside the token (`:648-649`, `schemas.py`), and the dialog copies it verbatim (`CheckinQrDialog.vue:32-33`) - no `window.location`, no URL building in the frontend at all, so link and QR cannot diverge. When `ALLOWED_ORIGINS` is unset both endpoints answer the same 422.

Leftovers swept: `useAppointments.ts` no longer sends `?origin=` (it was already ignored server-side) with the comment corrected; `render_checkin_qr_base64` is deleted (zero callers, `base64` import went with it); `permissions.md` lists the mint and QR endpoints under `appointments.write` plus the unauthenticated consume endpoint; the module CLAUDE.md has a check-in section; the `fix(review)` CHANGELOG entry is folded into the `feat` entry with its facts kept (secret key, attribution, limits, server-built URL).

Tests: the happy path asserts the minted `url` equals the allowlist origin plus `/p/check-in/<token>`, and `test_mint_rejects_unconfigured_allowlist` pins the 422. Suite is 11 passed; `dev-verify.sh --fast` 9 passed; eslint and ruff clean. Same typecheck caveat as the last two PRs: the worktree env reports pre-existing tree-wide resolution noise (nothing in the touched files), CI has the verdict.

Noted and out of scope here: the date guard (a natural part of the #480 time-window constraint) and the kiosk design discussion - I will weigh in on #480 separately.

ZoliQua (approved):
Approving — both of my round-3 points are done, and the link fix went further than needed.

- **The dead core page is gone.** `frontend/app/**` in the diff is now only the `'/p/check-in'` entry in `publicRoutes`, which is the precedent (as I got wrong and corrected in round 3). `git grep 'check-in?t='` on the head finds nothing.
- **The secret is wired.** `docker-compose.prod.yml:36` passes `AGENDA_PUBLIC_SECRET_KEY` next to the budget key, so a production deploy on that file can actually set it.

Ramón's blocker — the copied link pointing at the deleted page — is fixed at the right layer: `_checkin_base_url()` is the single source for both the minted `url` and the QR, the dialog copies what the server returns, and the frontend builds no URLs at all. With one origin source the link and the QR can no longer disagree, which was the real bug underneath the typo.

Locally on a fresh DB: `tests/test_appointment_checkin.py` = 11 passed, including the new allowlist-422 test. CI is green except `backend-test`, which is still running as I write. Agreed that the date guard belongs with #480.

ZoliQua:
@lamanji — the red `frontend-test` on run `35359023895` (the `useListQuery … expected undefined to be 'ana'` failure) was mine, not yours: a flaky test I added in #479 that the merge ref pulled into your run. Your later run went green on its own. The fix is #483. Sorry for the scare.

martinezsalmeron:
Merged — thanks @lamanji, and thanks @ZoliQua for the approval pass.

Re-verified on the merge result against a running stack: the dialog's "copy link" now yields `<ALLOWED_ORIGINS[0]>/p/check-in/<token>` (no query string), opening it lands on the public page and completes the check-in (`scheduled → checked_in`, event `note="qr-checkin"`, `changed_by` NULL), a replay is a 200 with no second event, and a tampered signature is a 401.

One small visual follow-up, no rush: in the **week view**, the hover container at `AppointmentCalendar.vue:816` sits on top of a name row that only reserves room for one icon (`ps-5`). With the QR button next to the dropdown trigger there are now two, so on hover the second icon covers the first letters of the patient name. Either widen the reserved padding or (my preference — one less always-rendered button per card) move "QR check-in" into the existing dropdown as a menu item.

Date guard and the kiosk shape stay with #480.

Issue #269, opened 2026-08-24
New module: prescriptions — clinical prescriptions (recetas) with per-country compliance hooks


martinezsalmeron:
From discussion #252. A base clinical prescriptions module, country-agnostic, with legal/format requirements per country shipped as localization modules on top — same pattern as billing → verifactu (ES) / india_gst (IN).

### Scope (base module)

Manifest: `depends: ["patients"]`, `category: official`, `installable=True`, `auto_install=False`, `removable=True` — but `uninstall()` **raises if any issued prescription exists** (clinical retention; verifactu precedent).

**Models** (own Alembic branch, e.g. `rx_`):
- `Prescription`: `clinic_id`, `patient_id` (FK, patients is in depends), `prescriber_id`, `status` (`draft | issued | cancelled`), `issued_at`, `notes`, `compliance_data` JSONB (per-country, keyed by country code — mirrors `Invoice.compliance_data`), `locale`.
- `PrescriptionItem`: `medication_name` (free text, **always** available), `catalog_ref` nullable UUID **without FK** (soft link to medication_catalog #223 — autofill when installed, keeps working when not; `medical_reference` pattern), `dosage`, `route`, `frequency`, `duration`, `instructions`, `sort_order`.
- `PrescriptionTemplate`: clinic-level favorite combos ("amoxicillin 500 mg / 8 h / 7 days + ibuprofen") stored as JSONB items; one-click fill on the create form. CRUD under Settings → Clinical (next to the treatment catalog, like #223).

**Lifecycle:** issued prescriptions are **immutable** — corrections are cancel + reissue (clinical-legal artifact, same rule as invoices). Never hard-delete.

**Prescriber identity:** per-professional `license_number` (rendered with a country-neutral label; country modules override it via `label_overrides` — "Cédula Profesional" MX, "Nº de colegiado" ES) and an optional uploaded **signature image** (stored via `media`). Open question: these live on the core professional profile vs. a module-owned per-user table — needs-design.

**PDF:** own `pdf.py` (WeasyPrint, following `billing/pdf.py`'s escaped structured contract). Clinic letterhead (logo, name, address — same clinic settings billing/budget use), date, patient, prescriber + license + signature image, items table. Download/print only in v1 — **no email delivery** (MX pharmacies reject emailed prescriptions; #252). Note: this is the third WeasyPrint copy after billing and budget — extracting a shared letterhead/PDF helper is a reasonable follow-up, not a blocker.

[...]

PR #465, opened 2026-09-15
feat(prescriptions): clinical prescriptions with compliance hooks (issue #269)

lamanji:
## Summary

- 4 commit(s): `feat(prescriptions): clinical prescriptions with compliance hooks (issue #269)`; `Merge remote-tracking branch 'upstream/main' into pr/prescriptions-module`; `fix(prescriptions): resolve layer-relative imports to composables (TS2307 on current main)`; `fix(prescriptions): normalize nullable item fields for UInput and indent permissions entry`
- **Scope:** modules `patient_timeline, prescriptions` | areas `backend, frontend, i18n, infra, testing`.
- **Why:** feat(prescriptions): clinical prescriptions with compliance hooks (issue #269)

## Tasks done

- `feat(prescriptions): clinical prescriptions with compliance hooks (issue #269)` - patient_timeline, prescriptions
- `Merge remote-tracking branch 'upstream/main' into pr/prescriptions-module` - core
- `fix(prescriptions): resolve layer-relative imports to composables (TS2307 on current main)` - prescriptions
- `fix(prescriptions): normalize nullable item fields for UInput and indent permissions entry` - prescriptions

## Modules touched

- `patient_timeline`
- `prescriptions`

## Checklist

- [ ] module `CHANGELOG.md` + docs updated per CLAUDE.md checklists
- [ ] catalogs + `frontend/modules.json` regenerated

## Test plan

- `scripts/dev-verify.sh --with-pytest --module patient_timeline prescriptions --fast` (compose `db` up)

## Refs

- #269

## Red-conditions (L76)

- [x] `backend/tests/modules/prescriptions/test_prescriptions.py::test_issue_empty_is_422`: red without the issue-gate in `PrescriptionService.issue` because an empty line set would be accepted and issued instead of rejected with 422
- [x] `backend/tests/modules/prescriptions/test_prescriptions.py::test_templates_crud_and_409`: red without the unique-name conflict check in `create_template` because a duplicate template would insert twice instead of answering 409
- [x] `backend/tests/modules/prescriptions/test_prescriptions.py::test_foreign_clinic_is_invisible`: red without clinic_id-scoped `issue` because a draft from another clinic would be issued instead of answered 404
- [x] `backend/tests/modules/prescriptions/test_uninstall_roundtrip.py::test_prescriptions_uninstall_roundtrip_is_branch_scoped`: red without the module's own Alembic branch (branch_labels + depends_on pat_0003) because uninstall would drop revisions outside `prescriptions`' chain

## Review


ZoliQua (changes requested):
Thanks for this — it is a substantial, carefully built module and the parts that are right are right: own tables and Alembic branch with a round-trip, permissions declared and enforced, tenant filtering on every query, a real state machine behind issuing, and documentation that exists rather than being promised. Four things to fix, then a design point for the lead.

1. **CI is red on our own module-boundary guard.** `test_no_new_cross_module_imports` fails with: `prescriptions/service.py imports app.modules.medical_reference` and `prescriptions/service.py imports app.modules.patients_clinical`. Either declare both in the manifest's `depends` — which is the honest answer if the module genuinely needs them — or route through whatever the depending module exposes. That is the one thing blocking mechanically.

2. **Prescriptions never reach the audit journal.** `activity_journal` records only the event types in its `_SUBSCRIBED` list, and neither `PRESCRIPTION_ISSUED` nor `PRESCRIPTION_CANCELLED` is in it, although both are published inside the transaction with `db=db` and therefore qualify under its own stated criterion. The patient timeline picks them up, but that is a patient-facing narrative, not the append-only log. For a document a pharmacist acts on, issuing and cancelling belong in the immutable record. Two lines.

3. **The Spanish manual claims something the code does not do.** The English screen doc says issuing freezes the prescription with a prescriber **snapshot** attached; the Spanish says "**firma** del prescriptor incluida" — signature included. The module renders no signature at all. That is a false statement about a legally meaningful element, in our primary UI language. Both manuals also describe filling a draft from a catalogue entry, which the screen does not offer.

4. **The agent tools roll back on a not-found.** Both tools catch `HTTPException` and call `ctx.db.rollback()`. The precedent in `agenda/tools.py` rolls back only on `IntegrityError`, to keep the session usable after a failed insert, and deliberately does not re-roll back when the service already did. A 404 does not poison the session, and rolling back discards whatever else the agent did earlier in the same turn. Please return an error result without the rollback.

Smaller: `last_verified_commit: HEAD` in both screen docs should be a real commit; a new domain term ("prescription", "receta") is not in `docs/glossary.md`, which the new-module checklist asks for; and the PR reorders `sdi_it`/`sistema_ts` inside `alembic.ini` while appending its own line — set equality holds, nothing is dropped, but it is noise on a lead-reviewed file, so please make it a pure append.

For @martinezsalmeron rather than for you: the compliance posture. Our accepted ADRs say e-prescribing stays out of scope — 0032 keeps E-Rezept with the approved PVS, 0033 notes that a Polish e-recepta for certain substance classes has no paper fallback at all, and 0035 Decision 10 says EPCS needs a per-version third-party audit. I read this module as printing a paper prescription and making no e-prescribing claim, which is consistent with all three, and I want to be explicit that I checked rather than assume. What does need his eyes is the `PrescriptionComplianceHook` registry and where prescriber identity lives: the new-module checklist asks for an ADR when an architectural decision is made, and that is one.

CODEOWNERS covers `backend/app/core/**`, `alembic.ini` and `frontend/app/**` here, so his review is required regardless.

ZoliQua (dismissed):
All four points are fixed. Verified on `38ab0928`:

1. **The boundary guard is green.** `manifest.depends` now declares `patients_clinical` and `medical_reference`, and `service.py` is untouched — you took the honest option rather than papering over the import. `tests/test_module_isolation.py` passes locally, `test_no_new_cross_module_imports` included.
2. **The journal subscribes.** `PRESCRIPTION_ISSUED` and `PRESCRIPTION_CANCELLED` are in `activity_journal`'s `_SUBSCRIBED`, with the `db=db` justification written next to them, and both catalogs are regenerated.
3. **The manuals agree with the code now.** ES reads "instantánea del prescriptor adjunta: nombre y licencia, **sin firma**" against EN's "no signature", and the template fill both describe is real — `PrescriptionTemplatesSettingsPage.vue` plus the template picker on the page. `last_verified_commit` is a real sha in both.
4. **No rollback on 404.** Both tools return the error without touching the session, with the `agenda/tools.py` precedent cited in the comment.

`alembic.ini` is a pure append now — the `sdi_it`/`sistema_ts` reordering is gone, the line differs from main by exactly `:app/modules/prescriptions/migrations/versions`. The migration declares `branch_labels = ("prescriptions",)` and `depends_on = ("pat_0003",)` for its `patients.id` FK, and no FK crosses into either newly declared dependency.

Locally on a clean database: `tests/modules/prescriptions/` + `test_module_isolation.py` + `test_manifest_validator.py` = 16 passed. (The one failure I saw is my own machine — a stale `__pycache__` tree from another branch trips the `version_locations` guard; that is #472, not this PR.) CI green, branch mergeable.

**Approving.** Two small things I would like in before it merges, neither worth another round on its own:

* **The glossary entry is still missing.** `docs/glossary.md` has no `prescription` / `receta` row, and the new-module checklist in CLAUDE.md asks for one on every new domain term. It is the one item from my "smaller" list that did not get done. Two lines.
* **`medical_reference` is `auto_install: False`.** Declaring it in `depends` is right, but it changes the install path: an admin installing prescriptions from the module admin UI now has to install the medication reference first, and neither `docs/technical/prescriptions/overview.md` nor the user manual says so. A sentence in the overview would save someone a confusing failure. (The CLAUDE.md note covers the developer side already.)

@martinezsalmeron — the design point from my first review is still open and it is yours: `PrescriptionComplianceHook` is a new registry in the same shape as `BillingComplianceHook`, and where prescriber identity lives is an architectural decision, so the checklist asks for an ADR. My compliance read stands: this module prints a paper prescription and makes no e-prescribing claim, which is consistent with ADR 0032 (E-Rezept via the approved PVS), 0033 (e-recepta) and 0035 Decision 10 (EPCS audit) — I checked rather than assumed. CODEOWNERS gives you `backend/app/core/**`, `alembic.ini` and `frontend/app/**` here in any case.

martinezsalmeron:
Hi @lamanji — thank you so much for this one. Prescriptions is a module people have been asking for since #252, and the foundations here are genuinely solid: own tables and Alembic branch, `clinic_id` on every query, events published inside the transaction, a real draft → issued → cancelled state machine, no hard deletes, and docs that actually exist. I installed it locally on `38ab0928` (it pulled `medical_reference` in automatically, so no confusing failure there — nice) and the install/uninstall round-trip was clean. Thanks also to @ZoliQua for the careful two rounds on isolation, the journal and the tools — all of that holds up.

I went through it from the angle of "a dentist prints this and hands it to a patient", and I found a handful of things I'd love to see sorted before we merge. None of them are big on their own, and I'd much rather land this well than fast. The first three I reproduced at runtime; the rest are from reading the code.

**Things that would bite a clinic**

1. **Archived allergies show up in the safety banner.** `prescribe_warnings` selects `Allergy.name` without `Allergy.status == "active"` (the soft-delete from #386). I added an archived allergy to a demo patient and it came straight back in the warnings. One extra `where` clause.
2. **Interaction flags render as a raw Pydantic repr.** `[str(flag) for flag in service_flags]` on a `PatientFlag` gives `type='interaction' risk_note='Riesgo de sangrado' involved=['Warfarina', 'Ibuprofeno']`, and that string goes into the `UAlert`. Something like `f"{' + '.join(flag.involved)}: {flag.risk_note}"` would read well.
3. **The PDF is always in English.** "Date / Patient / Prescriber / Medication / Dose / License no." are hard-coded and `prescription.locale` is never read. Billing and budget PDFs became locale-aware in #423 / #450 — following the same approach would be perfect. The letterhead also only prints the city; the issue asked for logo + name + full address (the same clinic settings billing uses).
4. **There is no way to enter the license number.** The `prescriber-profile` endpoints are there, but no screen calls them, so every PDF prints `Prescriber: … (License no.: )` with an empty value. Since the first user is in MX, where the cédula is mandatory, this is the one that matters most. A small form (even just a field on the prescriptions page or a settings section) is plenty for v1.
5. **Hook wiring in the PDF route.** `download_prescription_pdf` calls `hook.on_prescription_issued(row, db)` on every download, so a country module's issue-time side effects would re-run each time someone prints. I think the intent was to read the stored `row.compliance_data[country]` and pass it through `hook.enhance_pdf_data(...)`, which is currently never called (same for `get_required_fields`). No hook exists yet so nothing breaks today, but `prescriptions_mx` will be built on this contract, so it's worth getting right now.
6. **The PDF route serves drafts and cancelled prescriptions unmarked.** The UI only shows the button for issued ones, but the endpoint doesn't check. Either a 409 for non-issued, or a visible "DRAFT" / "CANCELLED" mark — whichever you find simpler.

**Frontend polish**

- The nav entry `/prescriptions` lands on "pick a patient" with no patient picker, so it's a dead end. Simplest fix is to drop the nav item and keep the patient-summary card + action as the entry points (or add a picker if you prefer).
- `save`, `issueRx` and `cancelRx` have no `try/catch` or toast — issuing an empty draft returns 422 and the user sees nothing. Issue and cancel are irreversible, so a small confirm step would be kind to users too.
- `can('prescriptions.write')` / `'prescriptions.issue'` are hard-coded; you already added the constants to `permissions.ts`, they just need using (`PERMISSIONS.prescriptions.write`).
- PDF download uses a plain `<a href>`; with an expired access token that opens a 401 JSON in a new tab. `api.raw()` handles the refresh (#451).

**Small ones**

- `docs/glossary.md` still needs the `prescription` / `receta` row (ZoliQua's note).
- `create_prescription_draft` always creates an empty draft (`items=[]`, `locale="es"`) — accepting items would make it useful to the copilot. Both tools also map every `HTTPException` to `not_found`.
- The uninstall gate only looks at `status == "issued"`; a prescription that was issued and then cancelled is still a clinical record (`issued_at IS NOT NULL` would cover both).
- `PrescriberProfile`'s docstring still says `depends == ["patients"]`.

**One thing that's on me, not you:** the hard `depends` on `medical_reference`. The issue asked for a soft integration, and declaring it was the honest way to satisfy the import guard — I don't think you did anything wrong there. But it does mean an `official` module now requires a `community` one, and the leftover `try/except ImportError` + `is_active()` guards are now dead code. Please don't spend time on this yet; I owe you a decision here together with the ADR on the hook registry and prescriber identity that ZoliQua flagged, and I'll follow up on #269.

Thanks again for the patience and for the steady stream of quality work — this is close, and I'm looking forward to merging it. Happy to clarify any of the points above. 🙏

lamanji:
Everything is in `c07d18a6` (plus stamp `9e80b48d`). Thank you for the thorough pass - point by point:

1. Archived allergies. `prescribe_warnings` filters `Allergy.status == "active"` (`service.py:341`); `test_archived_allergies_excluded_from_warnings` pins it.
2. Flag rendering. `_format_flag` produces `Warfarina + Ibuprofeno: Riesgo de sangrado` (`:369`), covered by a unit test.
3. Locale PDFs. Captions follow `prescription.locale` (`router.py:215`) with es/en sets and English fallback, the billing approach; letterhead is name plus the full formatted address (billing `_format_address` mirror). One honest gap: there is no `Clinic.logo` field anywhere in the codebase, so a logo cannot print yet - name and full address are in.
4. License entry. New prescriber-identity settings section (clinical, order 47) with a license field wired to the existing GET/PUT profile endpoints - the MX cedula path works end to end now.
5. Hook wiring. The PDF route reads the stored `compliance_data[country]` snapshot and passes the built payload through `enhance_pdf_data` (`router.py:200-218`), so issue-time side effects never re-run on download. `get_required_fields` is enforced at issue (`service.py:184-189`, 422 listing the missing fields).
6. Draft/cancelled PDFs. Visible DRAFT/CANCELLED mark, not a 409 - a cancelled copy stays printable for the record.

Polish: nav entry dropped (summary card plus New-prescription action are the entries); save/issue/cancel have `fail()` toasts with a confirm step on the irreversible two; `PERMISSIONS.*` constants everywhere; PDF downloads via `api.raw()`.

Small ones: glossary receta row; draft tools accept items with 404/409/422 mapped (not_found/conflict/invalid); uninstall gate is `issued_at IS NOT NULL` with direct guard tests; PrescriberProfile docstring fixed. The `medical_reference` depends question is untouched, awaiting your decision per your note.

Verified: module suite 12 passed plus the uninstall roundtrip, `dev-verify.sh --fast` 9 passed, eslint and ruff clean. Same typecheck caveat as the other PRs: the worktree env reports pre-existing tree-wide resolution noise (nothing in these files), CI has the verdict.

martinezsalmeron:
Hi @lamanji — merged. Thank you for this round: every single point from my last review is addressed, and this time I went through it in the browser as well as in the code.

**What I verified on `c07d18a6` + current `main`**

- Archived allergies stay out of the safety banner, and interaction flags read as plain text (`Warfarina + Ibuprofeno: …`).
- Full flow on a demo patient: draft → save as template → edit → issue → cancel → PDF. All of it works, no console errors, and both events land in the patient timeline and the activity journal.
- The PDF follows the prescription locale, prints the clinic name and full address, snapshots the license number at issue time, and a cancelled copy carries a clear CANCELADA mark. Fair point on the logo — there is no `Clinic.logo` to print yet, so that one is not on you.
- The new prescriber-identity settings section saves and is picked up by the next issued prescription.
- Hook wiring reads the stored `compliance_data` snapshot and goes through `enhance_pdf_data`; `get_required_fields` is enforced at issue.
- Module suite + isolation + manifest + journal tests: 32 passed on the merge with `main`.

One housekeeping note: `main` had moved (staff_attendance, treasury, QR check-in), so I pushed a merge commit to your branch (`46e6fa9d`) to resolve four mechanical conflicts — `alembic.ini` (pure append kept), the journal `_SUBSCRIBED` list, its events doc, and the regenerated catalogs. Nothing of yours was changed.

**Small follow-ups, whenever you have a moment**

None of these blocked the merge — they are the things I noticed while actually using the screen. A single small follow-up PR would be perfect, and please feel free to push back on any you disagree with:

1. **The UI never sends `locale`.** `createDraft` omits it, so every prescription is stored as `es` and the PDF is always Spanish — I had the UI in Portuguese and still got a Spanish PDF. Passing the current i18n locale in `createDraft` should be a one-liner.
2. **Issuing without a license number is silent.** I issued one before filling the profile and only found out from the PDF (`Licencia n.º: ` empty) — and since issued rows are immutable, the only way out is cancel + redo. A gentle warning in the issue confirm dialog when the profile has no license would save people from that. (A warning, not a block — country hooks already own the hard requirement.)
3. **The page does not say whose prescriptions these are.** There is no patient name and no link back to the patient record, so after landing from the summary card you lose your context.
4. **Editor details:** "Save" accepts a draft with zero lines; `&new=1` stays in the URL, so every reload reopens the editor; and there is no input for `route`, so the "Vía" column in the PDF is always empty (either add the field or drop the column).
5. **PDF polish:** a title ("Receta médica" / "Prescription"), a signature line for the hand signature, skip the "Notas:" row when there are no notes, and a localized date instead of ISO.
6. **Wording:** the PDF says "Licencia n.º" while the UI says "Cédula profesional". For the built-in `es` set I would use "N.º de colegiado" in both places (Spain's term) and leave "Cédula profesional" to the future `prescriptions_mx` hook via `label_overrides`. The two Spanish confirm strings are also missing the opening "¿".
7. **Tiny one:** while the confirm modal closes after "Issue", its title/text flip to the "Cancel" variant for the length of the animation, because `pendingConfirm` is already `null`. Keeping the last `kind` around until the modal is closed would fix it.

**Still on me, not you:** the hard `depends` on `medical_reference`, the ADR for the `PrescriptionComplianceHook` registry and prescriber identity, and how this module should sit next to the `documents` module's prescription type. Please don't spend time on any of those — I will follow up on #269.

Thanks again for the patience across three rounds, and to @ZoliQua for the careful reviews on isolation, the journal and the tools. This is a module clinics have been asking for, and it landed in good shape. 🙏

PR #466, opened 2026-09-15
feat(core): opt-in Sentry error tracking, DSN-gated, PII-free (local-only)

lamanji:
## Summary

- 4 commit(s): `feat(core): opt-in Sentry error tracking, DSN-gated, PII-free (local-only)`; `fix(core): cover sentry init happy path; lock sentry-sdk tree`; `Merge remote-tracking branch 'upstream/main' into pr/core-error-tracking`; `style(config): restore indentation lost in merge`
- **Scope:** modules `backend, testing` | areas `backend, testing`.
- **Why:** feat(core): opt-in Sentry error tracking, DSN-gated, PII-free (local-only)

## Tasks done

- `feat(core): opt-in Sentry error tracking, DSN-gated, PII-free (local-only)` - core
- `fix(core): cover sentry init happy path; lock sentry-sdk tree` - core
- `Merge remote-tracking branch 'upstream/main' into pr/core-error-tracking` - core
- `style(config): restore indentation lost in merge` - core

## Modules touched

- none (core/infra change; areas: backend, testing)

## Checklist

## Test plan

## Red-conditions (L76)

- [x] no guard-adjacent tests detected in this diff.

## Review


ZoliQua (changes requested):
Thanks — the gate itself is done right: no DSN means nothing initialises, a missing SDK warns and carries on, an init failure never breaks boot, `send_default_pii=False` is the correct default, and returning a bool makes the helper testable. Four things, and the first two are the ones that matter.

1. **`docs/technical/compliance-posture.md` needs to know about this, and I would treat that as blocking rather than a docs nit.** That file is where an operator goes to find out what leaves their deployment: §1 maps every category of regulated data, there is a dedicated paragraph for the one third-party path we ship today (the copilot redaction gate), and §6 lists the operator's own duties. This adds a second outbound sink and the document does not mention it. The sentence you wrote in the root `CHANGELOG.md` — that a GDPR clinic needs a DPA before pointing this at a cloud backend because breadcrumbs and URLs can carry identifiers — is exactly the right sentence; it just needs to live in the compliance posture too, with a row in the data map and an entry under operator duties.

2. **I believe the init runs too late to catch request exceptions.** `setup_error_tracking(...)` is called inside the `lifespan` body at `main.py:57`, while `app = FastAPI(...)` is at line 116 and the `add_middleware` calls at 129 and 143. Sentry's Starlette integration works by patching `build_middleware_stack`, and Starlette builds that stack on its first `__call__`, which is the lifespan message itself — so the ASGI middleware would never be inserted, unhandled request exceptions would not be captured, and only the logging integration would still report, which makes the feature look alive while missing exactly what it is for. Sentry's own guidance is to initialise as early as possible; moving the call to module scope above `app = FastAPI(...)` should settle it. I have not run this against a real DSN, so please confirm with a deliberate unhandled 500 before we merge — if I am wrong I would rather know than guess.

3. **Operators cannot switch it on.** Neither `SENTRY_DSN` nor `SENTRY_TRACES_SAMPLE_RATE` is in `.env.example` or `.env.prod.example`, and `docker-compose.coolify.yml` does not pass the DSN, so a Coolify deployment would need the compose file edited. `BUDGET_PUBLIC_SECRET_KEY` is the pattern to copy, including the comment explaining what it is for.

4. **On what actually leaves the building.** `send_default_pii=False` does suppress request bodies, cookies and client IP, but it does not touch the request URL — and the URL is where our identifiers live: `/api/v1/patients/{uuid}`, `/api/v1/budget/public/budgets/{token}/…`, and `?t=<token>` if #464 lands with the token in the query string. Please add a `before_send` and `before_send_transaction` that drop query strings and redact the token segment of the public routes, and set `environment` and `release` while you are in there. Related: `SENTRY_TRACES_SAMPLE_RATE` defaults to `0.1`, which turns on performance tracing by default. That is a second data flow — span descriptions carry SQL statement text through the SQLAlchemy integration — and it is not mentioned in the title or the changelog. I would default it to `0.0` and let the operator opt in a second time.

Small: one line asserting `Settings().SENTRY_DSN == ""` would pin the off-by-default promise at the settings level, where a future edit could flip it. The `sentry-sdk` block in `pyproject.toml` starts its comment at column zero while its neighbours are indented. And it is worth saying in the description that there is deliberately no frontend half and no DSN reaches the browser, because that is the first question any reviewer will ask.

`backend/app/core/**` and `main.py` are CODEOWNERS paths, so @martinezsalmeron needs to see this regardless — and given it is a new third-party data sink, his call on the posture is the right gate anyway.

martinezsalmeron (approved):
Reviewed the gate, the four points from @ZoliQua, and merged `main` (CHANGELOG conflict only). Pushed the follow-ups to this branch as c11fed74 rather than wait a round-trip, since they are all small and clearly specified.

**Verified, not guessed** — ran a deliberate unhandled 500 through a FastAPI app + the real `setup_error_tracking` inside the backend container, with a capturing transport:

- **Point 2 (init too late): not a problem.** sentry-sdk patches `Starlette.__call__` at class level (`integrations/starlette.py`, `patch_asgi_app`), so the SentryAsgiMiddleware wraps every request no matter when `init()` ran. Init in `lifespan` captured the `RuntimeError` with full request context, byte-for-byte identical to module-scope init. Left the call where it is.
- **Point 4 (URL leak): confirmed.** The event carried `query_string: "t=secrettoken"` and the patient UUID in `request.url`. Cookies were already `[Filtered]`. Also `environment` defaulted to `"production"` and `release` was empty.

**Fixed in c11fed74**

1. `scrub_event` as `before_send` + `before_send_transaction`: drops the query string, replaces UUID and long opaque path segments with `[id]`. After the fix the same 500 reports `url: ".../api/v1/patients/[id]"`, no query string.
2. `SENTRY_TRACES_SAMPLE_RATE` default `0.0` — tracing is a second opt-in.
3. `environment=settings.ENVIRONMENT`, `release=app.version`.
4. `SENTRY_DSN` in `.env.example`, `.env.prod.example`, `docker-compose.prod.yml`, `docker-compose.coolify.yml`.
5. `compliance-posture.md`: row in the §1 data map, a "shipped, opt-in" paragraph next to the copilot one (what still travels: stack traces, log text, breadcrumbs), operator duty in §6 (DPA before a cloud sink; keep log lines free of identifiers).
6. Tests: `Settings().SENTRY_DSN == ""` pinned, scrub unit test, init kwargs assert `before_send`. `pyproject.toml` comment indent.

No frontend half, no DSN reaches the browser — now stated in the changelog and the posture doc.

ZoliQua (changes requested):
Two of the four are in, and done well. `before_send` and `before_send_transaction` both run `scrub_event`, `environment` and `release` are set, and `SENTRY_TRACES_SAMPLE_RATE` now defaults to `0.0` so tracing is a second, deliberate opt-in. I ran your `_OPAQUE_SEGMENT` pattern against real URLs from this codebase and it does what it claims:

```
/api/v1/patients/b0eebc99-…-6bb9bd380a22          -> /api/v1/patients/[id]
/api/v1/patients/b0eebc99-…/documents             -> /api/v1/patients/[id]/documents
/api/v1/budget/public/budgets/AbCdEf01…/pdf       -> /api/v1/budget/public/budgets/[id]/pdf
/api/v1/billing/invoices/fb00bc99-…/pdf?locale=en -> /api/v1/billing/invoices/[id]/pdf?locale=en
```

Two things still open from the last round, and one new observation.

1. **The init still runs inside `lifespan`.** `setup_error_tracking(...)` is at `main.py:57`, `app = FastAPI(...)` at line 118. This is the point I care most about, because if I am right the feature silently misses exactly what it exists for. Moving the call to module scope above the `FastAPI(...)` construction costs nothing; if you would rather keep it where it is, please confirm with a deliberate unhandled 500 against a real DSN that the event actually arrives, and say so in the PR — I will take a measurement over my reading any day.

2. **Still nothing in the docs or the environment files.** `docs/technical/compliance-posture.md` does not know about the new sink, and neither `.env.example` nor `docker-compose.coolify.yml` mentions `SENTRY_DSN`, so a Coolify operator cannot switch it on without editing the compose file. The compliance one is the one I would call blocking: that document is where an operator looks to find out what leaves their deployment, and this is now the second outbound path.

3. **New, small:** `scrub_event` pops `query_string` and rewrites `url`, but the regex only matches a token that follows a `/`, so a token in a query value survives inside `url` if any integration ever puts the full URL there. Belt and braces would be to cut `url` at the first `?` as well — one line, and it makes the guarantee independent of how the integration splits the URL.

`backend/app/core/**` and `main.py` are CODEOWNERS paths, so @martinezsalmeron is needed here regardless.

ZoliQua:
Correction on my review — **I was wrong about the init order**, and I owe you the measurement that shows it.

I claimed the `setup_error_tracking()` call inside the `lifespan` body runs too late for Sentry's Starlette integration to see requests, so only log-level events would be reported. I asked you to confirm against a real DSN before merging; I should have done that myself first. I have now, against merged `main` (`dd2611ce`, clean tree):

* local Sentry-protocol ingest on `127.0.0.1:9999`, backend started with `SENTRY_DSN='http://abc123@127.0.0.1:9999/1'`
* one genuine unhandled 500 — `GET /api/v1/billing/invoices/<uuid>/pdf?locale=en&secret_token=shouldbescrubbed` (WeasyPrint has no libpango on this box, so the handler really does raise)

Result: an envelope goes out for that request.

```
HIT kinds=['event'] url='http://127.0.0.1:8000/api/v1/billing/invoices/[id]/pdf'
  "level":"error"
  "transaction":"/api/v1/billing/invoices/{invoice_id}/pdf"
  "value":"unhandled errors in a TaskGroup"
```

So request errors are reported as shipped. The path runs through `@app.exception_handler(Exception)` → `logger.exception(...)` → the logging integration, which is enough for the stated purpose.

The same envelope also verifies the scrubbing claims in the posture doc, which is the part I could not check by reading alone:

* `secret_token=shouldbescrubbed` — **not present** anywhere in the payload; `query_string` is absent entirely
* the invoice UUID is `[id]` in `request.url`, and the `transaction` is the route template, not the concrete path

And the two other things I asked for are in: `docs/technical/compliance-posture.md` names the sink in the data-flow table and in the "Error tracking today" note, and `SENTRY_DSN` / `SENTRY_TRACES_SAMPLE_RATE` are in `.env.example`. Both good.

One leftover, small and optional: `docker-compose.coolify.yml` passes `SENTRY_DSN` through but not `SENTRY_TRACES_SAMPLE_RATE`, so on Coolify tracing can't be turned on without editing the compose file. Harmless as long as the default `0.0` is what we want there — worth a one-line addition next time that file is touched.

Sorry for the noise on an otherwise clean PR.

PR #470, opened 2026-09-16
feat(payments): improve Razorpay follow-ups

tresundios:
## Summary
- Improve Razorpay payment handling for the India payment flow.
- Add batch gateway-info loading for the payments list to reduce per-row requests.
- Enforce clinic currency checks before initiating a gateway payment.
- Use clinic-local dates in payment creation.
- Tighten Razorpay polling and refresh behavior to avoid stale or runaway background polling.
- Remove dead webhook signature code from the Razorpay adapter.

## Modules touched
- payments
- payment_gateways
- razorpay
- frontend payment collection UI

## Test plan
- [ ] Verify the payments list loads gateway info correctly.
- [ ] Verify payment creation uses the clinic’s local date.
- [ ] Verify currency mismatch is rejected before provider initiation.
- [ ] Verify Razorpay checkout polling stops after timeout or dismissal and resumes on refresh.


ZoliQua (changes requested):
Thank you — this is a careful, well-aimed follow-up and most of the list from #439 is done properly. The batch endpoint is exactly right: one clinic-filtered query with `selectinload` on the refunds, a hundred-id cap against a page size of twenty, manual rows excluded client-side, and a cross-clinic test that asserts the empty result. `get_locked` with `FOR UPDATE` plus the three refresh tests closes the race cleanly. The currency check landed on both paths, including the webhook one, which is the half that actually protects us. The poll cap and `ondismiss` behave, and two dialogs cannot double-poll. `clinicToday()` with its Tokyo and Lima fixtures is a nice piece of work. CI is green and the branch is on top of main.

Two things block, and I checked both on the head rather than from the description.

1. **`payment_date` is still UTC where it counts.** `payment_gateways/service.py:331` is unchanged: `payment_date=confirmation.captured_at.date()`. What the commit fixed is the manual-entry default in the frontend, which is welcome, but my note on #439 was about this line. A capture at 03:00 IST is 21:30 UTC the day before, so it books into the previous day's cash-up and the wrong GST period. Please resolve the clinic timezone server-side with the house helper (`agenda/tz.py`, `get_clinic_tz` / `safe_zone`) and convert before `.date()`, with a test at an Asia/Kolkata instant just after local midnight. Keep `clinicToday()` and its tests.

2. **The 500 branch of the webhook commits partial work.** `razorpay/router.py:143` calls `await db.commit()` before raising, and the comment says so outright: "any partial state change before the failure commits together with the error record". `confirm()` flushes a `Payment` and its allocations before it mutates the request, so a failure after that point persists a half-applied confirmation *and* answers 500, and Razorpay then retries against that state. The terminal-state short-circuit saves the common case, but by luck rather than design. Please `rollback()` first, re-read the settings row, record the webhook error and commit only that, then raise. The 401 branch is fine as written and its test is welcome.

Then four smaller ones:

3. **`ar.json` is the English file.** I diffed it: the only difference from `en.json` is the missing trailing newline, and it contains no Arabic characters at all. That passes the parity check while shipping untranslated RTL UI — every other module layer ships real Arabic. Same for `modeKeyMismatchLive` and `modeKeyMismatchTest`, which are English in de, es, fr, hu, it, pl, pt and ta.
4. **No `CHANGELOG.md` entries** for `payments`, `payment_gateways` or `razorpay`, all three of which this touches.
5. **The new endpoint is undocumented**: `POST /payments/gateway-info/batch` is missing from the `payment_gateways` CLAUDE.md endpoint table, from `docs/technical/payment_gateways/overview.md` and from `permissions.md`. And `razorpay/CLAUDE.md` still says "all nine host locales" — it is ten as of this PR.
6. **`verify_webhook_signature` is still declared on the `GatewayAdapter` Protocol** (`adapters/base.py`) while the one real adapter no longer implements it. Removing the duplicate was right; the declaration should go too, or the next provider inherits a misleading obligation.

Take or leave: `useGatewayInfoBatch.ts` keeps `cache`, `pending` and `fetchBatch` at module scope, which is shared across requests in the SSR bundle — only `onMounted` reaches it today, but an `import.meta.client` guard would make that safe by construction; and `flush()` could chunk at a hundred so a larger page size can never 422 the whole row set. The manifest `author` is still "DentalPin Core Team" in both modules, which was the second half of the ADR ask, and the ADR header renamed `Deciders:` to `Decider:` while the other twenty-nine use `Deciders:`.

One thing I would keep and just ask you to announce: the Live/Test key-prefix validation is not in the PR title or body, but it answers the question raised on #439, it checks the merged state on the backend as well as the client, and it is well built. Please mention it in the description and the changelog so it does not arrive unannounced.

`frontend/app/**` and `docs/adr/**` are CODEOWNERS paths, so @martinezsalmeron's review is needed on top of mine.

tresundios:
reopened the same updates in this branch 
**payment-gateway-india-followups**

Also, the @ZoliQua change request has been implemented in the [`payment-india-followups-and-feedback`](https://github.com/HealtheKit/dentalpin/tree/payment-india-followups-and-feedback) branch.

Thus closing this PR

Issue #459, opened 2026-09-13
odontogram, agenda: three Nuxt UI components used in templates do not exist (UDropdownMenuItem / Group / Separator)


ZoliQua:
Found while reviewing #364 — **not** caused by that PR, which does not touch these files.

`@nuxt/ui` ships `DropdownMenu` and `DropdownMenuContent`, and nothing else in that family. Listing the package's component directory gives exactly:

```
DropdownMenu.vue
DropdownMenuContent.vue
```

There is no `DropdownMenuItem`, no `DropdownMenuGroup`, no `DropdownMenuSeparator` — in Nuxt UI v3 the menu contents come from the `:items` prop, not from child components.

Two files use the components that do not exist:

- `backend/app/modules/odontogram/frontend/components/odontogram/TreatmentBar.vue` — the "Add to plan" dropdown, whose whole body is `<UDropdownMenuGroup>` / `<UDropdownMenuItem v-for …>` / `<UDropdownMenuSeparator>`.
- `backend/app/modules/agenda/frontend/components/clinical/AppointmentModal.vue`.

On a dev server the console says so on every render of the dental chart:

```
[Vue warn]: Failed to resolve component: UDropdownMenuItem
[Vue warn]: Failed to resolve component: UDropdownMenuGroup
[Vue warn]: Failed to resolve component: UDropdownMenuSeparator
  at <TreatmentBar …> at <OdontogramChart …> at <ClinicalTab …>
```

**What I verified**: the component list above, the two call sites, and the three warnings in the browser console on the dental chart.

**What I did not verify**: what the user actually sees. An unresolved component renders as an unknown element, so the expectation is that the plan list and the "Create new plan" entry are simply absent — which would mean a dentist cannot attach a charted treatment to a plan from the chart — but I ran out of a clean way to open that menu in the browser and did not want to assert it without seeing it. Whoever picks this up should start there: select a treatment on the chart, open **Add to plan**, and see whether the menu has any entries.

The fix in both files is to move to the `:items` API, or to use `DropdownMenuContent` with plain markup inside.

ZoliQua:
Correction to my own report: only **one** file is affected, not two.

[...]

PR #471, opened 2026-09-16
fix(odontogram): rebuild the plan dropdown on the supported items API (#459)

ZoliQua:
## Overview

`TreatmentBar`'s "Add to plan" menu is built from `UDropdownMenuItem`, `UDropdownMenuGroup` and `UDropdownMenuSeparator`. Nuxt UI ships neither — the package contains only `DropdownMenu` and `DropdownMenuContent`, and in v3 the menu body comes from the `:items` prop. So those four tags resolve to nothing, and the dev console logs three `Failed to resolve component` warnings on every render of the dental chart.

Rebuilt on `:items`, the shape every other dropdown in the codebase uses (`PatientStickyHeader`, `PatientPaymentsPanel`, `RecallRow`, and `AppointmentModal`, which was already correct). The plan options become `type: 'checkbox'` items — the check indicator on the selected plan comes for free and `onSelect` still fires, so selection stays a single handler — and "Create new plan" is a second group, which renders the separator implicitly. The active plan keeps its amber leading icon through `ui.itemLeadingIcon`. Net −13 lines.

**One thing worth knowing before you weigh this up.** I went to verify the menu in a browser and could not make it appear, because it cannot: `showPlanSelector` requires a `treatmentPlans` prop, and nothing passes one — `OdontogramChart` hands `TreatmentBar` `selected-plan-id`, `plan-context-title`, `patient-id`, `mode` and `disabled`, and no other caller renders the bar. So this block is unreachable in the shipped app and the fix changes no screen a user sees today. It stops the markup from being a trap for whoever wires the selector up, and it clears the console warnings.

Which raises the real question, and it is yours rather than mine: **should the plan selector be wired up, or removed?** The code, its locale keys and its styles have been carried along unused. If you want it live, the missing piece is passing the patient's plans down; if not, deleting the block is smaller than this PR. I am happy either way and will do whichever you prefer — I fixed it in place because that is the conservative option.

## Test plan

- `npx nuxt typecheck` with the module layers loaded: clean. eslint on the file: clean apart from the pre-existing `v-html` warning.
- Verified against the installed `@nuxt/ui`: its component directory contains `DropdownMenu.vue` and `DropdownMenuContent.vue` and nothing else in that family, and `DropdownMenuContent.vue` binds `@select="item.onSelect"` on `DropdownMenu.CheckboxItem`, which is what the checkbox items rely on.
- **Not verified in a browser**, for the reason above: the block does not render. If you would rather have it wired up in this PR I will do that and then drive it properly.

Refs #459


martinezsalmeron (approved):
Reviewed in depth and driven in a browser.

**Diagnosis confirmed.** `@nuxt/ui` 4.6.0 ships only `DropdownMenu` + `DropdownMenuContent`; the only caller of `TreatmentBar` is `OdontogramChart.vue` and it never passes `treatmentPlans`, so the block is unreachable today, as you said.

**Browser check (the part you couldn't do).** I temporarily wired a fake `:treatment-plans` list into `OdontogramChart` locally (not committed) and opened a plan detail: the menu renders with both groups and the separator, the active plan keeps its amber star, the check indicator lands on the selected plan, `update:selectedPlanId` and `createPlan` both fire, the menu closes on select, and the console has no `Failed to resolve component` warnings.

Code is the same `:items` shape as the rest of the codebase, scoped to the module, CHANGELOG updated, CI green. Merging.

**On your question — wire up or remove:** my recommendation is remove. The flow moved to a fixed plan context (Diagnosis → create plan → PlanDetailView), and with the selector forced on it sits right under the "Adding to this plan: …" banner saying the same thing twice. Ramon has the final call; if he agrees, a follow-up PR deleting the block, the `treatmentPlans` prop, the two emits, the `.plan-selector` styles and the three locale keys would be welcome.

PR #472, opened 2026-09-16
test(core): skip source-less leftover dirs in the alembic version_locations guard

ZoliQua:
## What

`test_every_module_migrations_dir_is_registered_in_alembic_ini` walks every directory under `backend/app/modules/` and asks for an `alembic.ini` `version_locations` entry for each one that has `migrations/versions/`. It never checks that the directory is actually a module.

That matters because `git` does not remove ignored files on checkout. Switch from `main` to a PR branch that predates a module and the module's source disappears while its `__pycache__` directories stay:

```
backend/app/modules/patient_segments/__pycache__/service.cpython-313.pyc
backend/app/modules/patient_segments/migrations/versions/__pycache__/pseg_0001_initial.cpython-313.pyc
```

`migrations/versions/` still exists, so the guard fires — and on that branch `alembic.ini` has no entry for it, because the module was not merged yet:

```
AssertionError: Modules with migrations not registered in [alembic] version_locations of alembic.ini:
['app/modules/patient_segments/migrations/versions']. Append ':app/modules/<name>/migrations/versions' to the version_locations line.
```

The message asks you to register a module that is not on the branch. `patient_segments` landed in `cab5d344` (8 Sep), so today any branch forked before that hits this, and the state persists until you `git clean -xd`. I ran into it reviewing #456 and spent a few minutes convinced the PR had broken something.

## Change

Skip directories that are not packages — the same criterion discovery uses (`pkgutil.iter_modules` + `ispkg` in `loader._discover_filesystem`, which needs `__init__.py`). A leftover `__pycache__` tree has no `__init__.py`, a real module always does.

The walk moves into `_modules_missing_from_version_locations(modules_root, registered)` so the case is testable against a `tmp_path` tree instead of the repo: one package with migrations, one source-less leftover, and only the package is reported.

No behaviour change for a clean checkout — the guard still catches a module that ships migrations without an `alembic.ini` entry, which is what it exists for.

## Verified

`backend/tests/test_manifest_validator.py` — 9 passed, with the stale `patient_segments/__pycache__` tree still on disk here (before this change, 1 failed). `ruff check` and `ruff format --check` clean.


martinezsalmeron (approved):
Reviewed and reproduced locally — I happened to have the exact state on disk: `prescriptions/`, `staff_attendance/` and `treasury/` under `app/modules/`, nothing but `.pyc` files, left over from checking out #462/#463/#465 earlier today.

- On `main`: `test_every_module_migrations_dir_is_registered_in_alembic_ini` fails asking me to register three modules that are not on the branch.
- With this PR: 9 passed, leftovers still on disk.
- The guard still bites: removing `patient_segments` from `version_locations` on purpose fails it with the right message.
- The criterion matches discovery (`pkgutil.iter_modules` + `ispkg` in `loader._discover_filesystem`), and every real module with `migrations/versions/` has an `__init__.py`, so nothing legitimate drops out of the check.

Test-only, one file, helper extracted just enough to be testable against `tmp_path`. CI green. Merging.

PR #474, opened 2026-09-17
feat(payments): improve Razorpay follow-ups with feedback update

tresundios:
## Summary

- Address follow-up feedback from [#439](https://github.com/dentalpin/dentalpin/pull/439), [#445](https://github.com/dentalpin/dentalpin/pull/445), and [@ZoliQua](https://github.com/ZoliQua).
- Batch payment gateway-info loading for the payments list and add refresh locking to avoid payment-confirmation races.
- Enforce gateway and confirmed-payment currency checks, use clinic-local payment dates, and roll back partial webhook work before recording errors.
- Improve Razorpay checkout polling, Live/Test key-prefix validation, Arabic/localized messages, documentation, changelogs, and module metadata.

## Modules touched

- `payments`
- `payment_gateways`
- `razorpay`
- Frontend payment collection and Razorpay settings UI
- Documentation, ADRs, locales, and changelogs

## Test plan

- [ ] Verify gateway-info batch loading is clinic-scoped, excludes manual payments, and supports chunked requests.
- [ ] Verify refresh and webhook processing cannot confirm the same request concurrently.
- [ ] Verify gateway and confirmation currency mismatches are rejected.
- [ ] Verify captured payments use the clinic-local calendar date.
- [ ] Verify a failed webhook rolls back partial confirmation state before recording its error.
- [ ] Verify polling stops on timeout or dismissal and does not duplicate across dialogs.
- [ ] Verify Live/Test Razorpay key-prefix validation works in both frontend and backend.
- [ ] Run backend lint/tests and frontend lint/typecheck.

## Review

`frontend/app/**` and `docs/adr/**` are CODEOWNERS paths; review from [@martinezsalmeron](https://github.com/martinezsalmeron) is required in addition to payment review.


tresundios:
- Address the review feedback from [PR #470](https://github.com/dentalpin/dentalpin/pull/470), including [@ZoliQua](https://github.com/ZoliQua)’s requested changes. The implementation has been completed in the current [`payment-india-followups-and-feedback`](https://github.com/HealtheKit/dentalpin/tree/payment-india-followups-and-feedback) branch, including:
  - Booking captured payments using the clinic’s local calendar date instead of the UTC date.
  - Rolling back partially confirmed payment data when webhook processing fails.
  - Adding Arabic translations and correcting localized payment messages.
  - Documenting the gateway-info batch endpoint and updating related module documentation.
  - Removing the unused `verify_webhook_signature` requirement from the gateway adapter protocol.
  - Adding SSR-safe batching and request chunking to the frontend gateway-info composable.
  - Updating module metadata, ADR information, and changelog entries.

ZoliQua (changes requested):
Thanks for reopening this cleanly, and for the comments — this is one of the better-annotated diffs I have read here. Several of the changes explain *why* in a way that will still make sense to someone in a year, which is worth more than the code.

What I verified on `52b6d44e`:

- **Batch gateway info.** `get_for_payments_batch` is one query with `selectinload(refund_requests)`, clinic-scoped, and a missing id is simply absent — same contract as `get_for_payment` returning `None`. `payment_ids` is capped at 100 and `MAX_BATCH_SIZE` in `useGatewayInfoBatch.ts` mirrors it, so a page-size bump chunks instead of 422-ing every visible row. The dataloader is careful in the places that usually go wrong: the fetcher is captured inside `setup()` so the timer never calls `useApi()` without a Vue instance, and the module-scoped queue is client-gated so SSR cannot leak one request's cache into another's.
- **Refresh locking.** `get_locked` + `with_for_update()` on the manual refresh, so it serializes against a webhook landing on the same request. `from_model` touches only scalar columns, so dropping the `selectinload` that `get()` carries does not cost you a lazy load in the response.
- **Currency.** `supported_currencies` on the adapter contract, checked in `create_and_initiate` before the idempotency key is reserved, plus the confirmed-vs-requested check in `confirm()` — which matters because `record_payment` is only ever told `request.currency`.
- **Clinic-local payment date.** `captured_at.astimezone(clinic_tz).date()` with naive values defended as UTC, and `clinicToday()` on the frontend reading the Y-M-D fields off the wall-clock Date instead of re-shifting through `toISOString()`. The 00:15 IST case is exactly right.
- **Key prefix check against merged state**, not the incoming payload — a `mode`-only PATCH is still validated against the stored `key_id`. Good catch.
- **Dead code removed.** `verify_webhook_signature` was never called by the router; ADR 0029 updated to match.
- Razorpay locales are at full parity: all ten at 85/85 keys, Arabic included.

Locally: `tests/modules/payment_gateways/` + `tests/modules/razorpay/` + `tests/modules/payments/` = 103 passed, including the razorpay uninstall round-trip. CI is green except `backend-test`, still running as I write this.

### One thing to change

**The `author` field.** `payment_gateways/__init__.py` and `razorpay/__init__.py` both go from `"DentalPin Core Team"` to `"tresundios"`. I checked the history before saying anything: you did write both modules (`0a6ec7b9`, merged as #445), so the claim is factually supported and I would not be surprised if Ramón grants it. But it is a project-governance decision about two `category: official` manifests, it is unrelated to the follow-ups this PR is for, and it is not mentioned in the description. Please revert it here and raise it separately so it gets decided on its merits rather than landing inside a 1,400-line diff.

### One thing I want to understand before I approve

**The two webhook failure branches treat the same risk differently.** The `except Exception` branch now rolls back before recording, with a good explanation. The `except GatewayError` branch above it does not: it records the error and returns `{"ok": True}`, and `get_db()` commits on a successful return — so whatever is pending in the session at that moment is persisted.

Today that is not nothing. `apply_webhook_event` writes `request.raw_status_snapshot` (`last_event_type`, `last_event_id`) *before* dispatching, so a rejected event's snapshot is committed. I suspect that is deliberate and desirable — you want to see that the event arrived even though it was refused — but it does mean the branch is not write-free, and its safety rests on an unstated invariant: *every* `GatewayError` must be raised before any other write. I traced the current paths and it holds — `confirm()` does amount, currency and transition checks first, and the only `GatewayError` from `record_payment` is the `PaymentWorkflowError` wrap, which can only fire in the allocation validation that runs before the first `db.add`. So this is not a bug report.

It is fragile, though: one post-flush validation added to `record_payment` later, and the deterministic branch silently commits a half-written Payment with no `request.payment_id` and a 200 that tells Razorpay never to retry. Either make the branches symmetric (roll back, re-fetch, record, commit — the code is already written one block down), or state the invariant in a comment next to the `except GatewayError` so the next person changing `record_payment` knows it is load-bearing. Your call which; I would rather not guess what you intended.

### Two design points for @martinezsalmeron

1. **The CSP now names a vendor domain in core.** `frontend/server/middleware/csp.ts` hardcodes `https://checkout.razorpay.com` into `script-src`, `connect-src` and `frame-src` for every deployment, including the ~all of them that will never install the razorpay module. It is the first external origin any module has needed there, so it sets the precedent. The alternative — a `csp` key in the module manifest, or deriving the origin set from installed modules at request time — is more machinery than this one domain deserves today, but it is the difference between "core knows about one vendor" and "core knows about every vendor we ever add". Worth deciding now while there is exactly one.

2. **Resolving the clinic timezone is now written three times.** `agenda/tz.py` has `safe_zone`/`get_clinic_tz`, this PR copies it into `payment_gateways/service.py` (with an honest comment explaining why it is not cross-imported), and #462 adds a third copy in `staff_attendance`. All three read `Clinic.timezone`, which is a **core** model — so a small `app/core/clinic_tz.py` would be a legitimate shared helper rather than a boundary violation, and would stop the fourth copy getting the DST handling subtly wrong. `backend/app/core/**` is yours.

Everything else here I am happy with.

tresundios:
@ZoliQua I have updated the authors in the respective modules and Implemented the recommended symmetric webhook error-handling flow.  Both `GatewayError` and unexpected `Exception` paths now:

1. Roll back pending database changes.
2. Re-fetch fresh Razorpay settings.
3. Restore the rejected payment event snapshot.
4. Record the webhook error.
5. Commit only the receipt/error state.

The existing response contract is preserved: `GatewayError` returns `200`, while unexpected exceptions retain their existing failure response.

ZoliQua (approved):
Both points settled, and the webhook one is settled better than I asked for. Verified on `b4bd7a45`:

**`author`** is back to `"DentalPin Core Team"` in both manifests, with the catalog regenerated. Thank you — raise it as its own issue whenever you like; the history backs your claim and it deserves a straight answer rather than being bundled.

**The two failure branches are symmetric now.** I asked for either symmetry or a comment stating the invariant, and you took the harder option. What I checked in `_rollback_and_record_webhook_error`:

- `event = None` is bound before the `try`, so a payload `parse_webhook_event` rejects can no longer leave it unbound in either handler. That was a latent `NameError` in the old `except Exception` path too, not just in the new shared one.
- Everything written after the rollback is re-fetched — `settings` via `get_settings`, the request via `get_locked_by_provider_reference` — so nothing touches an expired instance.
- `record_webhook_received` is re-applied, which it has to be: it was flushed before the `try` and the rollback takes it with everything else.
- The `raw_status_snapshot` restore is scoped correctly: payment-type events only, guarded on `provider_reference`, skipped for `refund_*` which never write that field. And the docstring says *why* it is restored rather than just that it is — that it is audit data about the rejected event, not part of the failed confirmation. That is the distinction I was fishing for.
- The response contract is unchanged: `GatewayError` still answers 200 so Razorpay stops retrying, unexpected exceptions still 500 so it retries.

`test_gateway_error_after_confirm_rolls_back_payment_and_persists_error` is the right test for it, and it is a better test than the fix: it forces a `GatewayError` *after* `confirm()` has flushed a Payment, then asserts the payment count is 0, the request is still `awaiting_customer_action` with `payment_id is None`, the snapshot survived with the rejected event's id, and the error was recorded — and then retries the same webhook and asserts exactly one payment exists. That covers the half-applied-state failure end to end, not just the happy path of the new helper.

Locally on a clean database: 102 passed across `payment_gateways` / `razorpay` / `payments`, plus both uninstall round-trips on a virgin DB. CI is fully green and the branch is mergeable.

**Approving.**

The two design points from my last review stay open, but neither is yours and neither should hold this PR: the `checkout.razorpay.com` origin in the core CSP middleware, and whether clinic-timezone resolution should move into `app/core/` now that it exists in three places. @martinezsalmeron — those are the CODEOWNERS calls here (`frontend/app/**` is touched by the `clinicToday` wiring in `CollectAmountModal.vue`).

Issue #475, opened 2026-09-17
gdpr: employee personal data has no export or erasure path (payroll, attendance, activity traces)


ZoliQua:
## The gap

Our `gdpr` module is patient-only, by construction:

- `gdpr/models.py` — `GdprConsent.patient_id`, `GdprRequest.patient_id`, `GdprErasureAudit.patient_id` are all FKs to `patients.id`, and each model holds a `relationship(Patient)`.
- `gdpr/service.py` imports the patient model directly and builds the export / erasure payload from patient-owned tables.
- There is no plugin surface: no hook, no registry, no event another module can subscribe to in order to contribute rows to an export or an erasure. (`BillingComplianceHook` is the pattern we already use elsewhere for exactly this shape of problem.)

Meanwhile several modules already store **employee** personal data, which is in scope for the GDPR just as much as patient data:

- `payroll` — salary profiles, payslips.
- `staff_tasks` — who did what, when.
- `activity_journal` — per-user action traces.
- `staff_attendance` (#462, in review) — working-time records: who was in the building and for how long. This one is the sharpest case: an attendance log is the textbook example of employee monitoring data, with its own retention expectations, and the module ships no delete route and no retention rule.

So today an employee's subject-access request cannot be served by the product, and an erasure request cannot be executed. That is a real compliance gap, not a theoretical one: Article 15 does not distinguish between a patient and an employee.

## Why it is filed separately

Raised while reviewing #462; agreed there that fixing it inside a feature PR would be wrong. This needs a design decision first, and it touches `backend/app/core/**` if it becomes a registry.

## What needs deciding

1. **Data subjects beyond patients.** Does `gdpr` grow a subject type (patient / employee / contact), or does employee data get its own module? A subject type means the three `patient_id` FKs become nullable + a discriminator, which is a migration on a compliance table.
2. **A contribution surface.** Something like `get_gdpr_contributions()` on `BaseModule`, or a registry a module registers into on activate, so `payroll` / `staff_attendance` / `activity_journal` can each declare "these are my rows for this subject, this is how they export, this is how they erase". Modules must not be imported directly by `gdpr` — that would invert the dependency rule.
3. **Erasure vs. legal retention.** Payroll and attendance records normally have a statutory retention period that *overrides* an erasure request (in ES, working-time records are retained 4 years). So the answer to an employee erasure is often "retained under legal obligation", and the product has to be able to say that rather than silently delete or silently refuse.
[...]

PR #476, opened 2026-09-17
feat: Add Inbound Lead module

Valeriooooh:
## Summary

This PR adds a inbound leads module that can be useful for businesses that have a form where patients fill out their information to ask for appointments. It integrates directly with the patients and recalls workflow. When a form is submitted the Leads page gets a new entry with the patients data, the staff can subsequently make the call/email for the reservation, and add the new patient. 
If the patient fills in the form with a already present phone number/email, the module adds the patient to the recalls page for the staff to contact.

### Warning
AI was used to create this module, I reviewed the majority of the code myself, but this being my first time making a module for Dentalpin can make me overlook something.  

## Modules touched

- `leads` — new module (backend + Nuxt layer), own Alembic branch with two revisions
- `patients` — read and created through the existing `PatientService.create_patient` (no code change)
- `recalls` — call-backs created through the existing `RecallService.create` (no code change)
- host frontend: `frontend/app/config/permissions.ts`, `frontend/modules.json`
- registrations/config: `backend/pyproject.toml`, `backend/alembic.ini`, `backend/alembic/env.py`, `backend/app/config.py`, `.env.example`, `.env.prod.example`
- docs: `docs/technical/leads/*`, `docs/user-manual/{en,es}/leads/*`, `docs/glossary.md`, `docs/modules-catalog.md`

### Always

- [x] PR title follows Conventional Commits — suggested: `feat(leads): inbound enquiry queue with recall routing`
- [x] `cd backend && ruff check . && ruff format --check .` passes (whole backend)
- [x] `cd frontend && npm run lint` passes (0 errors; the 6 warnings are pre-existing, in other modules)
- [x] Tests added or updated when behaviour changed — 44 backend tests + 1 marked round-trip, 27 new frontend unit tests
- [x] No cross-module imports outside `manifest.depends` — only `patients` and `recalls`, both declared

### If a new module was added

- [x] Entry point registered in `backend/pyproject.toml` (`leads = "app.modules.leads:LeadsModule"`)
- [x] `backend/app/modules/leads/CLAUDE.md` written (purpose, public API, dependencies, permissions, tools, configuration, lifecycle, gotchas)
- [x] `backend/app/modules/leads/CHANGELOG.md` started (`## Unreleased`)
- [x] Alembic migrations on the module's own branch — `leads_0001` (`branch_labels=("leads",)`, `down_revision="0001"`, `depends_on=("pat_0003",)`) and `leads_0002` (availability becomes structured; no branch label, same branch)
- [x] `backend/scripts/generate_catalogs.py` re-run — `docs/modules-catalog.md` updated; `docs/events-catalog.md` unchanged
- [x] `removable=True` → uninstall round-trip test added (`backend/tests/modules/leads/test_uninstall_roundtrip.py`), downgrading `leads@-2` and asserting the `recalls` tables survive
- [x] `frontend/modules.json` regenerated with the canonical command (`node scripts/modules-json.mjs ../backend/app/modules --prefix /module_layers`)

- [x] ~~New event added to `backend/app/core/events/types.py`~~ — n/a: this module publishes and consumes **no events of its own**. The recall it creates publishes `recall.created` through the `recalls` module's existing contract.
- [x] ~~`generate_catalogs.py` re-run~~ — run; event catalog unchanged
- [x] ~~Payload shape documented~~ — documented as "publishes nothing of its own; the recall it creates publishes `recall.created`" (module CLAUDE.md)

- [x] Returned by `get_permissions()`: `["read", "write", "settings.read", "settings.write"]` → `leads.read`, `leads.write`, `leads.settings.read`, `leads.settings.write`
- [x] Listed in `manifest.role_permissions` (admin `*`; dentist `read`; assistant + receptionist `read`/`write`; `settings.*` admin-only on purpose — rotating the key breaks the clinic's website until a human pastes the new one in)
- [x] Added to `frontend/app/config/permissions.ts` (`PERMISSIONS.leads.*`)
- [x] Endpoints protected with `require_permission("leads.…")`; the two cross-module routes also carry that module's permission (`POST /` → `patients.read`, because it returns the matched patient's name; `POST /{id}/convert` → `patients.write`)

- [x] Target module is in `manifest.depends` — `patients.id` (`ondelete="SET NULL"`), pinned by `depends_on=("pat_0003",)`. **No FK to `recalls`**: the recall is the recalls module's own row, linked by `(clinic_id, patient_id, reason)` and deduped there.
- [x] `manifest-consistency` green; the module documents that it cannot be installed without `recalls`

- [x] ~~ADR added~~ — n/a: no new architectural decision. It follows ADR 0001 (plugin contract), ADR 0002 (per-module branch, required by `removable=True`) and ADR 0005/0023 for permissions and session auth.

- [x] Appended to `docs/glossary.md` — *Lead / Solicitud entrante*, *matched enquiry → recall / consulta de paciente conocido → Recordatorios*, *availability days / días disponibles*, *availability slot / franja horaria*, *intake key*, *daily intake cap*, *honeypot*, *convert*

- [x] `backend/app/modules/leads/CHANGELOG.md` updated under `## Unreleased`
- [x] ~~Other modules' CHANGELOGs~~ — n/a: no existing module's code changed (`patients` and `recalls` are consumed, not modified)
- [x] User manual: `docs/user-manual/{en,es}/leads/` (index + 2 screens, EN/ES sets identical) with the frontmatter contract
- [x] `docs/technical/leads/overview.md` + `permissions.md`

## Test plan

- [x] **Backend**
  - `cd backend && python -m pytest tests/modules/leads -q` → **44 passed** (routing matrix, byte-identical intake bodies, daily cap, tenancy, permission split, structured availability, `extra="forbid"`, tool-surface guard)
  - `python -m pytest -m alembic_roundtrip tests/modules/leads -q` → **1 passed** (branch-scoped downgrade/upgrade over both revisions; `recalls`, `recall_contact_attempts`, `recall_settings` survive)
  - Module gates green: `test_module_docs.py`, `test_entry_point_parity.py`, `test_module_manifests_consistency.py`, `test_alembic_branches.py`, `test_module_service.py`, `test_cli_modules.py`, `test_frontend_layers.py`, `test_generate_catalogs.py`
- [x] **Frontend**
   - `cd frontend && npx vitest run` → **26 files / 189 tests passed** (27 new: week-strip/availability rules, form validation, name-split/phone helpers; plus the locale-parity guard over 10 locales × 89 keys)
  - `npx nuxt typecheck` (host + every module layer) → exit 0
  - `npx eslint . module_layers` → 0 errors
- [x] **Docs gates** — `check_docs_coverage.py --strict` PASS, `check_docs_layout.py` OK
- [ ] **Reviewer: end-to-end check** (needs the stack; note that a local `pytest` run drops every table in the dev database, so run `./scripts/reset-db.sh && ./scripts/seed-demo.sh` afterwards)
  1. `docker compose exec -T backend python -m app.cli modules install leads` + restart (module is `auto_install=False`; `recalls` resolves transitively).
  2. **Settings → Integrations → Formulario web**: generate the intake key (shown **once**, with Copy buttons), set the daily limit to 2.
  3. Post the `curl` snippet from that page → `201 {"data":{"received":true}}`; a card appears on `/leads` (queue opens filtered to **New**, chip reads `Status · 1`).
  4. Post again with a phone that belongs to an existing patient → **identical 201 body**, no card, and a `high`-priority call-back due today appears in **Recalls** with the enquiry text and the availability as `mon, wed · afternoon`. Post it twice more → still one recall, note appended once.
  5. Availability: send `"availability_days":["thu","tue","thu"],"availability_slot":"afternoon"` → the strip highlights **Tuesday and Thursday** and *afternoons*, and the stored value is `["tue","thu"]` (mon-first, deduped). Send the retired `"availability":"Tardes"` → **422**.
  6. Wrong/disabled key → bare `401`; a third enquiry past a cap of 2 → `429` + `Retry-After`, gauge reads "3 of 2 today".
  7. Staff: *Nuevo contacto* with an existing patient's phone → toast says *"… is already a patient — a call-back has been added to Recalls"* (no card). With a fresh phone → card created; the week picker stores the days you tap.
  8. Click a card → drawer shows the enquiry panel (motive, description, week strip, phone, email, date) with the patient form pre-filled — and **notes empty**, on purpose. *Crear paciente* → patient created, card flips to `converted` and leaves the default view.
  9. Validation: type `marta@example` in the email field and tab out → the field is marked with a message in the UI language and the form refuses to send; same for a 3-digit phone or a future birth date.
  10. Log in as a receptionist → `/leads` works; **Settings → Integrations → Formulario web** is not reachable (403).

### Notes for reviewers

- **The daily cap is clinic data, not env config.** It lives in `leads_settings.daily_cap` (default 200, `0` = unlimited), edited on the settings page — deliberately **not** a `LEADS_INTAKE_DAILY_CAP` env var, so a clinic can raise its own ceiling mid-campaign. Only `LEADS_INTAKE_MAX_BODY_KB` and the optional captcha credentials are env vars.
- **Abuse controls are layered, and their limits are documented** in `docs/technical/leads/overview.md`: key → honeypot → body cap → per-clinic DB cap. Rate limits use the existing slowapi limiter (off outside production, per-process); the DB-backed cap is the real ceiling. A **captcha was considered and deliberately not shipped** — it is inert without external credentials, and operator-level spam belongs in front of the app (Cloudflare rate rules or nginx `limit_req`), not in this module. No `Caddyfile` change (stock Caddy has no `rate_limit` directive).
- **The intake response never varies by branch** (new lead / queued recall / matched opted-out patient all return the same 201 body): an unauthenticated caller must not be able to probe "is this phone one of your patients?".
- **The 429 path commits the counter before it raises.** `get_db` rolls back a session whose handler raised, so without the explicit `await db.commit()` the increment would be lost and a flood would look like it never happened. `test_intake.py` asserts `day_count` through a *separate* connection exactly for that.
- **Availability is structured and the intake contract changed with it**: day codes (`mon`..`sun`) plus an optional slot replaced free text, because the front desk reads a week strip, not a sentence. `leads_0002` drops the old column (free text cannot be parsed into days — guessing a call window is worse than asking again), and `extra="forbid"` on the lead payloads makes a stale website form fail **loudly** with a 422 instead of silently losing the value. Weekday labels come from `Intl.DateTimeFormat` in the UI (correct in all ten locales), so there are no per-day i18n keys; only `leads.slots.*` is translated.
- **Motive and availability never reach the patient record.** The convert drawer starts with empty notes; the only place they are rendered outside the module is the recall note (matched-patient path, where no lead card exists), where availability is a language-free token (`mon, wed · afternoon`) because a stored note cannot be localized.
- **Deliberate deviation from the original internal plan:** the `/leads` queue opens filtered to **New** (a *visible* chip, `Status · 1`), so converted cards leave the default view. The API still treats "no status param" as "every status", and a **Show all statuses** action in the empty state keeps the history one click away.
- **All ten UI locales are translated** (en, es, fr, de, pt, it, pl, hu, ta, ar — 90 keys each, key-parity and placeholder-checked by `frontend/tests/i18n/locale-parity.test.ts`). Each locale reuses its host locale's word for Recalls and its own address register. Note the `{'@'}` escape in `leads.validation.email`: a literal `@` is vue-i18n linked-message syntax and fails to compile the file otherwise.
- **Known gap, documented in the module CLAUDE.md:** a *repeat* enquiry from a `do_not_contact` patient stacks a second `needs_review` recall, because the dedupe only looks at *active* recalls. Fixing it means widening `ACTIVE_STATUSES` inside `recalls` — deliberately out of scope here.
- **Not covered by automated tests:** clicking through the drawer/modal/settings page in a real browser (the settings page is client-registered by a plugin, so it does not render under SSR). Template/type errors are caught by `vue-tsc` + ESLint; steps 4–10 above are the manual pass.


ZoliQua (changes requested):
Welcome, and thank you — both for the module and for saying up front that AI helped write it. That makes a reviewer's job easier, not harder: I know to read for the things a generator gets wrong rather than for typos.

I approved the CI run (first-time contributors need a maintainer to start it) after reading the diff. Locally on a fresh database: `tests/modules/leads/` + `test_module_isolation.py` = 46 passed, uninstall round-trip included. CI is green so far, with `backend-test` and `e2e` still running.

The public endpoint is the part I read most closely, and it is well built: the honeypot answers success and writes nothing; missing, unknown and inactive keys are one indistinguishable 401; the rate-limit bucket is the SHA-256 of the key, never the key itself; the response does not vary between "new lead" and "known patient" (your D12), so the form is not an oracle for "is this phone one of your patients"; and the daily-cap counter is committed *before* the 429, so a flood shows up in the gauge instead of vanishing with the rollback. Tenancy is clean — every query outside the key lookup filters on `clinic_id`, and the key lookup is correctly the one place that *derives* the clinic. `patients` and `recalls` are honestly in `depends`, and recalls are written only through `RecallService.create`, so dedupe and the `recall.created` event keep working. That is a better first module than most.

Two things I would like changed.

### 1. A stranger's words land on a real patient's call list, under the patient's name

When an enquiry's phone or email matches a patient, `route_matched_enquiry` queues a high-priority "call today" recall on that patient, with `motive` and `description` as the note. The *submitted* name, phone and email are dropped. I checked it against the service rather than assume — patient Juan Garcia, `+34 612 345 678`; a form submitted as "Somebody Else", `612345678`, `stranger@example.com`:

```
OUTCOME: recall_queued
RECALL on patient: Juan Garcia  priority=high due=2026-09-18
NOTE:
Cancel my implant surgery

I am allergic to penicillin, please note it.
--
submitted name in note: False
submitted email in note: False
```

The front desk sees Juan with an urgent note, apparently from Juan, asking to cancel his surgery and flagging an allergy. Nothing in the record says the text came from an unauthenticated web form filled in by someone else. It does not have to be malicious: a relative, a typo in one digit of the phone number, or two patients sharing a number all produce the same result.

The fix is small and keeps D12 intact, because the response to the submitter does not change: put the identity *as submitted* at the top of the composed note, e.g.

```
Web form — submitted as: Somebody Else · 612345678 · stranger@example.com
```

so whoever makes the call can see the mismatch before acting on the text. A test that routes a mismatched name and asserts it appears in `reason_note` would pin it.

### 2. `list_leads` hands unauthenticated free text to the cloud LLM path

`get_lead` is correctly `exposes_free_text=True`. `list_leads` is not, but it returns `_lead_summary(lead)` for every row, and `_lead_summary` already includes `motive` (200 chars) and `description` (up to 4,000 chars). So listing leads through the copilot sends free prose to the cloud model, which is exactly what the flag exists to prevent under redaction (CLAUDE.md, "New agent-exposed capability").

It matters more here than in most modules, because this text is written by *anyone on the internet* through the public form. That makes it a prompt-injection surface as well as a PHI one — "ignore previous instructions and…" in a description would reach the assistant's context the next time someone asks it to summarise the lead queue.

I think the original intent was to leave the free text out of the summary: `_get_lead` sets `data["description"] = lead.description` after calling `_lead_summary`, which would be pointless if the summary already had it. Dropping `motive` and `description` from `_lead_summary` restores that and keeps `list_leads` safe. Flagging `list_leads` with `exposes_free_text=True` is the alternative, but it would take the whole list tool off the cloud path.

### Smaller

- `route_matched_enquiry` uses `date.today()` for `due_date`, which is the server's calendar day, not the clinic's. Near midnight in a clinic whose timezone differs from the server's, "call today" lands on the wrong day. Several modules resolve `Clinic.timezone` for exactly this; see `agenda/tz.py`.
- The settings docs recommend a server-side POST, which is right. Worth one more sentence: if a clinic *does* post from a browser form, the key is in the page source and effectively public, and the rate limits and daily cap become the only protection. Clinics should hear that from us.
- `_enforce_body_limit` only checks the declared `Content-Length`, as your own comment says; a chunked request skips it. The field `max_length`s still bound what gets stored, so this is a note, not a request.

### For @martinezsalmeron

This one needs you anyway: `alembic.ini` and `frontend/app/config/permissions.ts` are CODEOWNERS paths, and it adds a new unauthenticated public endpoint. Three things for you rather than for Valerio:

- **Consent and retention for leads.** The form collects contact details and free-text health information from people who are not patients yet. There is no consent capture on intake and no retention rule, and discarded leads are kept indefinitely. `gdpr` is patient-only (#475 covers the same gap for employee data), so leads sit outside it until they convert. Probably a follow-up, but it should be a decision.
- **`docs/technical/leads-plan.md` is 1,752 lines.** The code cites its decision numbers (D8, D12), so it is load-bearing rather than a leftover. But a few of those decisions — no oracle on the public endpoint, routing known patients to recalls — are the kind the checklist puts in an ADR. Your call whether the plan stays in `docs/technical/` or gets distilled.
- `category: "official"` for a first community contribution — fine by me if it is fine by you.

Thanks again. The two fixes above are small, and the hard parts are already right.

Valeriooooh:
Hi, thank you for the quick response!
I completely overlooked the fact that someone could make "false" call requests in other person's name. I will update that part like you asked, I agree with you. Still on the call request, do you think that I should allow custom fields/customizable fields in the integration settings? I'm writing this module for a family clinic but I also want to make sure it can be useful for as much people as possible. The way I was envisioning this is keeping the Name,Phone number and email as required fields, but the rest like availability or time preference as customizable (sent in a json custom field) , since this fields will not be used to fill patient information.

I will also make the remaining changes you proposed.

Thanks again for the quick response and the good spirit!

ZoliQua:
Glad it was useful, and thank you for taking it so well.

On custom fields: my advice is **not in this PR**, and when it comes, as an admin-declared schema rather than a free JSON blob. Here is why, from what this endpoint already does well.

Right now `LeadIntakeCreate` is fully typed and `extra="forbid"`: an unknown field is a 422, and every field has a bound (`max_length=200/4000`, `DayOfWeek`, `AvailabilitySlot`). That strictness is part of your security model, not a limitation. A free-form `custom: {…}` on an **unauthenticated** endpoint would undo three things at once:

- **Storage abuse.** Anyone holding the key could post arbitrary nested JSON up to the body cap on every request, and nothing would validate its shape.
- **Personal data you cannot describe.** The clinic could no longer say what the form collects, and that is the first question under GDPR. Free-text health information arriving in untyped fields is the worst case of it.
- **The copilot point from my review.** Every custom value is more unauthenticated free text that could reach the agent's context.

It would also lose what your typed fields give you today: `availability_days` renders with localized weekday names on the lead card. A JSON value would render as whatever string the website sent.

If a second clinic asks for something concrete, the shape that keeps all of that is:

1. The admin **declares** the extra fields in the intake settings: key, label, type (`text` / `choice` / `boolean`), max length, required or not.
2. The public endpoint **validates** submissions against that declaration and still rejects unknown keys.
3. Values are stored as a bounded dict next to the typed columns, shown on the lead card with the admin's labels, and treated as free text wherever `exposes_free_text` matters. They never feed the patient record automatically.

So: keep name, phone, email, motive, description and availability exactly as they are, ship this PR with the two fixes, and open an issue for "admin-configurable intake fields" describing the family-clinic case that made you want it. That is also the right place for @martinezsalmeron to weigh in, since it is a product call as much as a technical one.

Valeriooooh:
That is a very good point! I will proceed with the changes.

ZoliQua:
Heads-up before you go looking: the red `frontend-test` is **not yours**. It is a flaky test I added in #479 (`useListQuery … expected undefined to be 'ana'`), and CI runs it against the merge ref even though your diff is backend-and-docs only. 193 of 194 frontend tests passed. The fix is #483, waiting to be merged. Sorry for the noise — that is the third PR it has hit.

I approved the workflow run for your new push, and I am going through the changes now.

ZoliQua (dismissed):
Approving. Both points are fixed, and the first one I re-measured rather than read: same scenario as before — patient Juan Garcia on `+34 612 345 678`, a form submitted as "Somebody Else" / `612345678` / `stranger@example.com`:

```
NOTE:
Formulario web — enviado como: Somebody Else · 612345678 · stranger@example.com

Cancel my implant surgery

I am allergic to penicillin, please note it.
```

The front desk now sees whose words those are before picking up the phone. You went further than I asked by writing the header in the *clinic's* communication language, with all ten locales covered and a test guarding the key set — and the reasoning in the comment (a stored note has no reader locale, so it must be frozen in the clinic's language) is the right call, not the easy one.

**The tool leak** is closed the way I hoped: `motive` and `description` are out of `_lead_summary`, so the four unflagged tools no longer carry enquirer prose, and only `get_lead` returns it, still flagged `exposes_free_text=True`. `test_tools.py` pins the shape, so adding a prose field back to the summary will fail a test rather than silently re-open it in five tools.

The three smaller items are in too: the recall's "call today" is now the clinic's local day rather than the server's, and the settings doc says plainly that a browser-side form puts the key in the page source and leaves the rate limits as the only protection.

Locally on a fresh database: `tests/modules/leads/` + `test_module_isolation.py` = 54 passed, uninstall round-trip included.

**On CI: the red `frontend-test` is mine, not yours** — the flaky `useListQuery` test from #479, which CI runs against the merge ref. 193 of 194 frontend tests passed; #483 fixes it. Everything else, including `backend-test` and `e2e`, is green.

Still for @martinezsalmeron before this can merge: `alembic.ini` and `frontend/app/config/permissions.ts` are CODEOWNERS paths, plus the three points from my first review that are his rather than Valerio's — consent and retention for lead data (`gdpr` is patient-only, same shape as #475), whether the 1,752-line `leads-plan.md` stays in `docs/technical/` or gets distilled into an ADR, and `category: "official"` for a first contribution.

Nice work, and thank you for taking the security points seriously rather than defensively.

Valeriooooh:
Thank you for the opportunity! 
I will keep in touch in the future for further improvements.

Valeriooooh:
I addressed the merge conflicts

martinezsalmeron:
Merged — thank you, Valerio, and thanks @ZoliQua for the two rounds.

I re-ran the whole thing on a fresh stack before merging: the leads test suite + module gates green locally, then the full manual pass in a browser — key generation, all intake branches by curl (new lead, matched patient with a stranger's name, repeat dedupe, retired `availability` → 422, wrong key → 401, honeypot), the queue, the convert drawer with validation, the manual "Novo pedido" routing toast, and the recall with its identity header on the call list. Everything behaved exactly as the PR description says.

Three small things I fixed directly on `main` in b184b38d rather than ask for another round:
- the convert drawer gated Edit/Create on `patients.write` only, so a dentist (`patients.*`, `leads.read`) could click into a 403 — now `leads.write` (+ `patients.write` for convert), mirroring the API;
- the edit modal's success toast showed the "Status" label as its message;
- manifest `category: community` / `author: Valeriooooh` — it is your module, not the core team's.

The rest is tracked: #500 (small UI follow-ups — yours if you want them, none urgent) and #499 (consent/retention for lead data, a product decision on my side).

PR #477, opened 2026-09-17
fix(alembic): join version_locations on literal colon for Windows

lamanji:
## Summary

- `backend/alembic/env.py` joins `version_locations` on a literal `:` instead of `os.pathsep`, matching the `version_path_separator = :` already declared in `backend/alembic.ini`.
- On Windows `os.pathsep` is `;`, which silently breaks Alembic branch-graph resolution (`alembic heads/history/upgrade`).
- Extracted from local `stash@{0}`; the ini half of that stash already landed upstream, so this PR carries only the still-missing `env.py` hunk.

## Modules touched

- none (core Alembic bootstrap only)

### Always

- [x] `cd backend && ruff check . && ruff format --check .` passes (verified on `alembic/env.py`)
- [x] `cd frontend && npm run lint` passes (no frontend changes)
- [x] Tests added or updated when behaviour changed (no behaviour change on Linux; Windows-only join fix)

## Test plan

- [x] `python -m alembic -c alembic.ini heads` resolves the full branch graph on a Windows host (previously broken by `;`-joined paths)
- [x] ruff check + format clean on the touched file


martinezsalmeron (approved):
Merging — the change is correct and zero-risk: `alembic.ini` declares `version_path_separator = :`, so `env.py` should join on the same literal, and the old comment (`= os`) was simply wrong. On Linux/macOS/Docker `":" == os.pathsep`, so this is byte-identical there.

One correction to the description, for the record, so nobody chases this later:

- This hunk does **not** fix `alembic heads/history/upgrade` on Windows. `heads` never executes `env.py`, and for the commands that do, the `ScriptDirectory` graph is already built from the **static** ini list before `env.py` runs (that is exactly what the M1 guard docstring in `tests/test_manifest_validator.py` says: "env.py's runtime discovery override is too late for graph building").
- Verified: forcing `os.pathsep = ";"` on main, `env.py` does override the option with `;`-joined paths, yet `heads` (50) and `current` (15) are identical to the `:` run. The real Windows fix was the ini half that already landed.
- Caveat: `discover_version_locations` returns absolute paths, so on Windows a `:` join would split `C:\...` at the drive letter if anything ever consumed the option. Nothing does today, so it is moot — but it means this override is effectively dead for in-tree modules. Candidate follow-up: either delete the override or make it actually feed graph building for out-of-tree modules (the use case the ini comment promises).

Thanks!

Issue #388, opened 2026-09-06
ci(frontend): jobs run `npm install`, so a lockfile out of sync with package.json passes CI and breaks `npm ci` in Dockerfile.prod


martinezsalmeron:
## What happened

PR #375 shipped a `frontend/package-lock.json` that had dropped the nested `node_modules/@nuxt/cli/node_modules/{cac@6.7.14,commander@13.1.0}` entries. Every CI job was green. But:

```
$ npm ci
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: cac@6.7.14 from lock file
npm error Missing: commander@13.1.0 from lock file
```

Reproduced on npm 10.8.2 (compose dev image) and 10.9.8 (`node:22-alpine`, the `Dockerfile.prod` base). That is the release-image path (`release.yml` → `Dockerfile.prod` → `npm ci`), so the next tag would have failed to build. Fixed by hand on the branch (3b1d3887) before merging.

## Why CI missed it

Every frontend job in `.github/workflows/ci.yml` installs with `npm install` (lines ~227, 248, 270, 292, 360). `npm install` silently repairs a drifted lock in the runner and never fails. Only `npm ci` validates sync.

The Python side already has this guard: `python-lock-freshness` fails on `uv.lock` drift (#371). npm has none.

## Fix

- Switch the frontend jobs to `npm ci --no-audit --no-fund`. It is also faster and deterministic.
- If any job legitimately needs `npm install` (e.g. the e2e job adds a package), keep it there and add one explicit `npm ci` step, or a `npm-lock-freshness` job mirroring the uv one.

Refs #371 (uv.lock guard), #375 (where it bit).

PR #478, opened 2026-09-17
chore(tooling): pin frontend node engines, enforce lf endings for sh/yml

lamanji:
## Summary

- `frontend/package.json` declares `engines: { node: "^20.19.0 || >=22.12.0" }`, matching the actual nuxt/vite/vite-builder constraints (a stale Node-20 local image caused a 30-hour SSR misdiagnosis).
- `.gitattributes` gains `*.sh text eol=lf` and `*.yml text eol=lf` (the repo already lost a backend to a CRLF entrypoint: `exec ... no such file or directory`).

## Modules touched

- none

### Always

- [x] `cd backend && ruff check . && ruff format --check .` passes (no backend changes)
- [x] `cd frontend && npm run lint` passes (no frontend code changes; `package.json` is JSON-validated)
- [x] Tests added or updated when behaviour changed (no behaviour change)

## Test plan

- [x] `python -c "json.load(open('frontend/package.json'))"` parses; `engines` value confirmed
- [x] `git status` after the `.gitattributes` change shows only the two intended files (no line-ending flip cascade)


martinezsalmeron (approved):
Merging — both halves check out:

- **`engines`**: `^20.19.0 || >=22.12.0` is byte-for-byte the constraint declared by `nuxt@4.4.2`, `vite@7.3.1` and `@nuxt/vite-builder@4.4.2`. No `engine-strict` anywhere, so this only warns; CI (`node-version: "22"`) and the pinned `node:22-alpine` image (v22.23.2) both satisfy it.
- **`.gitattributes`**: all 20 tracked `*.sh`/`*.yml`/`*.yaml` files are already LF in the index (`git ls-files --eol`), so no renormalization cascade. `backend/docker-entrypoint.sh` — the file behind the CRLF incident — is covered.

Two non-blocking nits for a follow-up:

1. **`package-lock.json` is now 3 lines out of sync.** npm mirrors `engines` into the lock's root package entry; regenerating with `npm install --package-lock-only` on this branch adds exactly that block and nothing else. `npm ci` does not validate it (CI and the prod-image e2e build are green on this head), so nothing breaks — but the next `npm install` anyone runs will dirty the lock, and given #388 we'd rather not carry drift on main.
2. **`bin/dentalpin`** is a bash script without a `.sh` extension, so it is still only under `text=auto` and would get CRLF on a Windows checkout with `core.autocrlf=true` — same failure mode as the entrypoint. One more line (`bin/dentalpin text eol=lf`) closes it.

(Minor: the description mentions `*.sh`/`*.yml`; the diff also adds `*.yaml`, which is fine.)

Thanks!

Issue #473, opened 2026-09-17
/patients always tries to get status active  Even when cleared


BabuBahir:
### What happened

http://localhost:3000/patients always tries to get status active , When cleared
http://localhost:3000/patients?status=active .

Also If we choose [archived](http://localhost:3000/patients?status=archived) It still tries to get Active 

### Steps to reproduce

Go to this page , and toggle the active/archived Status , Choose Archived and then click Clear 

<img width="1891" height="717" alt="Image" src="https://github.com/user-attachments/assets/f8b93513-910f-4f2b-87e2-b9faf6083c28" />

ZoliQua:
Thanks — reproduced both halves, and they turn out to be two different things.

**The clearing bug is real and now has a fix in #479.** `useListQuery` writes the filters into the URL and re-parses them back on every route change. An empty value serializes to `null` and the key was dropped from the URL entirely — but a URL cannot tell "empty" from "absent", while the parse side can, because a missing key falls back to the declared default. The patients list declares `status: ['active']`, so clearing the filter produced a URL with no `status`, which re-parsed straight back to `['active']`. I pinned it in a test before touching anything:

```
AssertionError: expected [ 'active' ] to deeply equal []
```

#479 writes an explicit empty marker when a value serializes to nothing but its default does not, which is the only case where the two states differ. Filters whose default is already empty (`q`, `city`, …) still leave the URL, so no page gains `?q=&city=` noise.

**The second half is not a bug in that code — it is a UI-vs-API mismatch**, and I would rather it were decided than patched. The backend has no status filter at all; it has `include_archived: bool` (`patients/router.py:63` → `service.py:178-179`, which only adds `Patient.status != 'archived'` when the flag is off). So the page maps `status` containing `archived` to `include_archived=true`, and that means *active **and** archived* — there is no archived-only query to send. The control is presented as a status picker but behaves as an include-archived toggle, which is why selecting only Archived still lists active patients.

Two ways out:

1. **Relabel the control** — make it a single "Include archived" switch, matching what the API actually does. No backend change, honest UI, but you can never look at archived patients alone.
2. **Add a real status filter** to `GET /api/v1/patients` (`status: list[str] | None`) and keep `include_archived` for compatibility. More work, but "show me archived patients" is a reasonable thing for a receptionist to want, and the UI already promises it.

@martinezsalmeron — option 2 touches the patients router and service; your call which one we want. I will leave this issue open for that part after #479 lands.

[...]

PR #479, opened 2026-09-17
fix(ui): keep a cleared list filter cleared when its default is not empty (#473)

ZoliQua:
Fixes the first half of #473.

## The bug

`useListQuery` serializes filters into the URL and re-parses them back on every `route.query` change. `serialize()` returns `null` for anything empty, and `pushUrl` skipped those keys:

```ts
const s = serialize(v as FilterPrimitive)
if (s !== null) next[k] = s
```

A URL cannot tell "empty" from "absent" — but the parse side can, because `parseInto` falls back to the declared default for a key that is not present. So for any filter whose default is *not* empty, clearing it is lossy:

1. user picks Archived → `?status=archived`
2. user clears the filter → `status` serializes to `null` → key dropped → `?` with no `status`
3. the `route.query` watcher re-parses → `parseInto` sees no `status` → falls back to `defaults.status` → `['active']`

The chip snaps back to Active and the list goes on requesting active patients, which is exactly what the issue describes.

Reproduced in a test before fixing — `filters.value.status` comes back as `['active']` after being set to `[]`:

```
AssertionError: expected [ 'active' ] to deeply equal []
```

Today `backend/app/modules/patients/frontend/pages/patients/index.vue` is the only list with a non-empty default (`status: ['active']`), so it is the only page affected — but the defect is in the shared composable, so the next list that declares one would inherit it.

## The fix

Write an explicit empty marker when a value serializes to nothing **but its default does not** — the only case where the two states differ:

```ts
if (s !== null) {
  next[k] = s
} else if (serialize(defaultsDict[k] as FilterPrimitive) !== null) {
  next[k] = ''
}
```

`parseInto` already handles `''` for every filter type it supports (`'' → []` for arrays, `'' → false` for booleans, `'' → ''` for strings), so the round trip closes without touching the parse side at all.

Filters whose default is already empty — `q`, `city`, `do_not_contact: null`, `with_debt: false` — are unchanged and still leave the URL, so no page gains `?q=&city=` noise. That is the second test case.

## Not in this PR

The issue's other half — "choose Archived and it still returns Active" — is not a bug in this code path. The backend exposes only `include_archived: bool` (`patients/router.py:63`, `service.py:178`), so `status: ['archived']` sends `include_archived=true`, which means *active **and** archived*; there is no archived-only mode to ask for. That is a UI-vs-API mismatch needing either a new backend filter or a relabel of the control, and it should be decided rather than patched. Commented on the issue.

## Verified

`frontend/tests/composables/useListQuery.test.ts` — 2 new tests, both failing before the change and passing after. Full frontend suite: 24 files, 164 tests passed. `npm run lint`: 0 errors (6 pre-existing warnings in module layers, none touched here).


ZoliQua:
Pushed a second commit — the composable fix alone was not enough, and would have made things quieter but not better.

With only the first commit, clearing the status filter stays cleared (the chip empties, the URL says `status=`), but `fetcher` only ever set `include_archived` when the selection *contained* `archived`:

```ts
if (q.filters.status.includes('archived')) params.set('include_archived', 'true')
```

So an empty selection still requested active-only patients while the UI showed no status filter at all. That is a worse failure than the original: the snap-back was at least visible, this would have been a filtered list under an unfiltered chip.

`fcccbc6f` treats an empty selection as *no status restriction* — which, given the API's only widening knob is `include_archived`, means sending it. It also makes the cleared state count towards the active-filter badge, because it genuinely changes the result set (archived patients appear).

The archived-only mismatch is still untouched and still needs the decision I described on the issue.

martinezsalmeron (approved):
Merging. Diagnosis is exact and the fix is the smallest one that closes the round trip — nice that it needed nothing on the parse side.

**Verified in the browser (dev stack, admin@demo.clinic):**

- On `main`: Archived → Clear leaves `/patients` and the chip snaps back to `Estado · 1` — the issue, reproduced. Bonus manifestation you didn't list: clearing straight from the default state empties the chip but the URL doesn't change (`/patients` → `/patients`), so no refetch happens at all and the list silently stays active-only. Your fix covers that too, since `?status=` is now a real URL change.
- With this branch: Clear → `/patients?status=`, chip stays empty, request goes out with `include_archived=true`; opening `/patients?status=` cold rehydrates the cleared state; re-selecting Active → `?status=active` and the request drops `include_archived`; mobile "Filtros (1)" counts the cleared state.
- No noise elsewhere: `/invoices` `?status=draft` → Clear → `/invoices`. Checked the other three consumers' `defaults` — all empty, so patients is indeed the only page affected today.
- No console errors. `closingIssuesReferences` is empty, so #473 stays open for the archived-only half as you intended.

**Non-blocking follow-ups** (fine to fold into whatever lands for the second half of #473, since that will rewrite the same text anyway):

1. The encoding table in the `useListQuery.ts` header (`string[] → omitted when empty`) and the "URL serialization rules" table in `docs/technical/lists-redesign.md` are now incomplete: empty is written as `?key=` when the declared default is non-empty.
2. Per the repo checklist, a behaviour change on a screen should touch `docs/user-manual/{en,es}/patients/screens/list.md` + bump `last_verified_commit`. That page is already stale on main (its troubleshooting entry talks about a "Show archived switch" that doesn't exist), so I'm not holding this fix for it — but it needs a line saying a cleared status filter includes archived patients.
3. Theoretical only: a `number` filter with a non-null default set to `null` would round-trip `'' → Number('') → 0`. No consumer has one; worth a guard only if one ever appears.

On the open decision in #473 (relabel vs real `status` filter) — that one is on me, I'll answer on the issue.

Thanks!

PR #483, opened 2026-09-18
test(ui): isolate the useListQuery tests so they stop fighting over the router

ZoliQua:
The `useListQuery` tests I added in #479 are flaky, and they have already turned two unrelated PRs red. This fixes them.

## What happened

Since #479 merged, `frontend-test` failed on 2 of 9 CI runs, both on the same assertion in the second case:

```
× useListQuery … > still omits a filter whose default is already empty
AssertionError: expected undefined to be 'ana'
```

- run `35341650972` on #460 (`feat/mcp_init`)
- run `35359023895` on #464 (`pr/agenda-qr-checkin`)

Neither PR touches the frontend. Both went red because CI tests the merge ref, which includes my test. Sorry to @BabuBahir and @lamanji for the noise.

## The cause was not the sleep

My first guess was the fixed 400 ms `settle()` against a 300 ms debounce plus an async navigation, so I swapped it for `vi.waitFor` polling. That was not enough. Under full CPU saturation locally (36 busy processes on 12 cores), the polled version still failed once in 20 runs, and it failed after polling for **3 s** with `q` still `undefined`. A slow runner cannot explain that; something was actively undoing the push.

That something was test 1. Everything in the file shares one Nuxt app and one router, and `mountSuspended` does not unmount between tests. So by the time test 2 runs there are two live `useListQuery` instances watching and pushing the same URL. Instance 2 pushes `?q=ana`, instance 1 pushes its own state without `q`, and whichever debounce lands last wins. When instance 1 wins, instance 2 re-parses `q = ''` and settles there for good.

## The fix

- Unmount the component after each test, and `router.replace({ query: {} })` before each.
- Keep `vi.waitFor` in place of fixed sleeps, since the URL update really is a debounce plus a navigation.

## Evidence

Same saturation, same machine:

| version | failures |
|---|---|
| test as merged in #479 | 1 / 12 |
| `waitFor` only | 1 / 20 |
| isolated + `waitFor` (this PR) | **0 / 40** |

It also still catches the bug it exists for: run against the composable from before #479, the #473 case fails with `expected 'active' to be ''`.

Full frontend suite: 24 files, 167 tests passed. `eslint` is clean on the file. No production code changes.


ZoliQua:
@martinezsalmeron — a nudge on this one, because the flake it fixes keeps costing other people.

Since #479 merged, `useListQuery.test.ts` has turned **three contributor PRs** red on a test unrelated to their diffs:

- #460 (@BabuBahir), run `35341650972`
- #464 (@lamanji), run `35359023895`
- #476 (@Valeriooooh), run `35479150760` — today, 193/194 frontend tests green, the one failure mine

Each time someone has to be told it is not their fault, and @lamanji spent a round writing his own fix for it in #484 before reverting it. This PR is test-only, all 12 checks are green, and the fix is measured: 0/40 failures under full CPU saturation against 1/12 for the version on `main`. Whenever you have a minute.

ZoliQua:
Rebased onto `main` now that `2b9793ed` has landed, so this PR is only the part your commit does not cover. Your change is the right first half — the fixed 400 ms sleep against a 300 ms debounce was indefensible, and polling removes it.

What is left is the reason the race exists at all: **every test in the file shares one Nuxt app, and therefore one router, and `mountSuspended` never unmounts.** By the second test there are two live `useListQuery` instances watching and pushing the same URL.

I demonstrated that directly rather than arguing from flake statistics — two instances mounted, the second sets `q=ana`, then the first pushes its own state:

```
URL after the first instance pushes: {"q":"ana","status":""}
second instance now sees q = "ana"
```

The first instance is still watching the route, still re-parses it into its own filters, and is free to overwrite what the second wrote. Whichever debounced push lands last wins, which is why the failure mode was `q` staying `undefined` **after a 3 s poll** — not a slow runner, a converged wrong state.

So this adds `afterEach(unmount)` and a `beforeEach` route reset on top of your polling, and keeps `vi.waitFor` exactly as you wrote it.

**One honest note on the evidence.** When I measured before, the poll-only variant still failed 1 in 20 runs under full CPU saturation, against 1 in 12 for the original. I re-ran the comparison today, 16 runs each of your version and this one, and got **0 failures on both** — at a few percent, that is what you would expect from a run that short, so today's numbers do not separate them. The deterministic demonstration above is the stronger argument, and it is why I think the isolation is worth having even though your fix already makes the failure rare.

If you would rather not carry a second change for this, closing it is fine by me — but then I would want the interference written down in a comment in that file, because the next person adding a test there will hit it.

lamanji:
@ZoliQua keep it - the deterministic two-instance demonstration settles it, and the interference deserves the fix rather than just a comment. My polling made the failure rare; your isolation removes the reason it exists at all.

And thank you for the attribution and the apology - none needed, but gracefully said. The `2b9793ed` landing was a nice surprise; glad the first half proved useful.

martinezsalmeron (approved):
Reviewed and verified locally.

Confirmed the interference and its actual trigger: `mountSuspended` itself does `await router.replace('/')` before the component's setup (`@nuxt/test-utils` runtime-utils), so mounting test 2 clears `?status=` while instance 1 is still watching. Instance 1 re-parses to the default and schedules a push of `status=active`; if that navigation commits before instance 2's debounced push fires, instance 2 re-parses `q=''` and its pending push becomes a no-op — `q` stays `undefined` for good. Deterministic repro on `main`: add a 100 ms wait between mount and `setFilter('q','ana')` → fails 100%. Same wait with this PR → passes.

8/8 runs of the file green, full suite 24 files / 167 tests green. Merging.

PR #484, opened 2026-09-18
feat(billing): invoice PDF labels for all UI locales (#422)

### LINKED ISSUE #422: billing: translate the invoice PDF labels into fr/pt/de/hu/pl/it/ar

lamanji:
## Summary

- 1 commit(s): `feat(billing): invoice PDF labels for all UI locales (#422)`
- **Scope:** modules `billing` | areas `backend, testing`.
- **Why:** feat(billing): invoice PDF labels for all UI locales (#422)

## Tasks done

- `feat(billing): invoice PDF labels for all UI locales (#422)` - billing

## Modules touched

- `billing`

## Checklist

- [x] catalogs + `frontend/modules.json` regenerated (no manifest/event changes; freshness gates green)

## Test plan

- `scripts/dev-verify.sh --with-pytest --module billing --fast` (compose `db` up)

## Refs

- #422

## Red-conditions (L76)

- [x] no guard-adjacent tests detected in this diff.

## Closes

- Closes #422

## Review


lamanji:
Re-pushed with one more commit on top of the labels (`5259d2ee`): the `frontend-test` failure on the first run was a timing flake in `useListQuery.test.ts`, unrelated to this PR (backend-only diff).

Evidence: byte-identical failure twice (`expected undefined to be 'ana'`, 23 files green around it) in a test whose `settle()` sleeps a fixed 400ms against the 300ms search debounce plus an async router round trip - about 100ms of margin that a loaded CI runner can eat. It is also the only fixed-sleep helper in the whole frontend suite.

The fix keeps every assertion identical and swaps the sleep for `vi.waitFor` on the asserted end state (5s timeout), so the test now polls instead of racing. Test-only change, no production code touched; flagging it here since the file is freshly yours - happy to split it out if you would rather land it separately.

ZoliQua (changes requested):
Thanks — closing #422 is worth doing, and the structure is right: one dict per locale in the same shape as `es`/`en`/`ta`, `dir="rtl"` only for `ar`, and a test per locale heading. I checked key parity mechanically across all ten dictionaries: every locale has the same keys and the same six `status` entries. `test_invoice_pdf_format.py` + `test_pdf_locales.py` = 9 passed locally, and CI is green.

Two things to change, both small.

### 1. `grand_total` is still English on three invoices

`grand_total` is the biggest label on the page, and three locales still carry the English placeholder:

| locale | now | should be |
|---|---|---|
| `de` | `TOTAL` | `GESAMTBETRAG` (or `GESAMT`) |
| `hu` | `TOTAL` | `VÉGÖSSZEG` |
| `pl` | `TOTAL` | `RAZEM DO ZAPŁATY` (or `RAZEM`) |

`fr`, `pt` and `es` are correct as `TOTAL` — it is the native word there — and `it` already has `TOTALE`. The heading test cannot catch this, since it only checks `invoice`.

### 2. Please drop `5259d2ee` from this PR

Thank you for chasing the flake — you were right that it was not your diff. But #483 already fixes it, and swapping the sleep for `vi.waitFor` is not enough on its own. I measured that exact variant under full CPU saturation, and it still failed once in 20 runs — after polling for 3 s, with `q` still `undefined`. A longer timeout cannot help when the state never converges.

The real cause is that the first test's component is never unmounted. Every test in the file shares one Nuxt router, so two `useListQuery` instances push competing URLs. #483 unmounts in `afterEach`, resets the route in `beforeEach` and keeps `waitFor`: 0/40 under the same load. If both PRs land, the second one conflicts on the same file. Keeping this PR to billing is cleaner, and it keeps your diff backend-only as the description says.

### Terminology — suggestions, your call

Two credit-note terms describe a *different kind* of document from the one `es` and `pl` already use. `es` says **Factura Rectificativa** and `pl` says **Faktura korygująca** — both mean "corrective invoice", which is what a `credit_note_for_id` document is here.

- **`hu`: `Jóváíró számla` → `Helyesbítő számla`.** "Jóváíró számla" is everyday speech. The legal term for a document that corrects an issued invoice is *számlát módosító okirat*, commonly *helyesbítő számla*. That is also what a Hungarian accountant will look for, and what `nav_online` reports as a `MODIFY` operation. With it, `credit_note_for` reads better as `Az eredeti számla` than `Jóváíró számla ehhez`.
- **`de`: `Stornorechnung` → `Rechnungskorrektur`.** *Stornorechnung* implies a full cancellation of the original invoice. A partial correction is a *Rechnungskorrektur*.

Smaller `hu` wording, only if you are editing anyway: `total_paid` reads more naturally as `Befizetett összeg`, and `balance_due` as `Fizetendő összeg`.

### One note on Arabic

I could not render the PDF here (WeasyPrint needs libpango, which this machine lacks), so I checked by reading: `fonts-noto-core` in the backend image includes the Noto Arabic faces, and Pango falls back per glyph, so the text should shape correctly. The CSS names `'Noto Sans Tamil'` explicitly, though. Adding `'Noto Sans Arabic'` the same way would make the Arabic rendering deterministic instead of depending on fontconfig's fallback order. A screenshot of an `ar` invoice from the running stack would settle the RTL table layout better than I can by reading.

lamanji:
@ZoliQua all addressed in `639de1dd` (on top of the revert `1ce099b2`).

1. Native grand totals: de GESAMTBETRAG, hu VÉGÖSSZEG, pl RAZEM DO ZAPŁATY. fr/pt/es stay TOTAL, it stays TOTALE. The heading test now has a sibling, `test_each_locale_renders_own_grand_total`, so the biggest label is pinned per locale too.
2. `5259d2ee` reverted - the flake is yours in #483 and this PR is billing-only again. Heads-up: CI on this push runs your original timing test once more, so it may flake a final time until #483 lands.
3. Terminology taken as suggested: hu Helyesbítő számla + Az eredeti számla, de Rechnungskorrektur (+ `-zu` on the reference), hu Befizetett összeg and Fizetendő összeg.
4. Arabic: `Noto Sans Arabic` added to the font stack next to Tamil. On the screenshot: I tried to render one locally and failed on my own toolchain (mapper bootstrap rabbit hole, then no reachable browser and no WeasyPrint/pango here - same limits you have). Best evidence I can offer is the test asserting `dir="rtl"` plus the right-aligned amount CSS that was already global. If the RTL table still looks off when you can render it, point me at the spot and I will fix it blind from the markup.

Verified: pdf-format suite 8 passed, ruff clean, backend battery gates green. The diff is `5259d2ee..639de1dd`.

ZoliQua (approved):
Approving — all four in, checked on the head rather than from the summary.

- `grand_total` is native now: `GESAMTBETRAG` / `VÉGÖSSZEG` / `RAZEM DO ZAPŁATY`, and `test_each_locale_renders_own_grand_total` pins it per locale, so the biggest label on the page can no longer silently stay English.
- The flake commit is reverted; the diff is billing-only again (3 files).
- Terminology: `Helyesbítő számla` / `Az eredeti számla`, `Rechnungskorrektur` / `Rechnungskorrektur zu`, and the two Hungarian wording tweaks. As a native Hungarian speaker I read the whole `hu` dictionary again after the change — it reads like a Hungarian invoice now.
- `'Noto Sans Arabic'` sits next to Tamil in the font stack.

I re-ran the key-parity check across all ten dictionaries (same keys, same six `status` entries), and `test_invoice_pdf_format.py` + `test_pdf_locales.py` = 10 passed locally. CI green. Agreed on the Arabic screenshot: neither of us can render it here, so the RTL table layout stays a follow-up for whoever runs the full stack first. Thanks for the quick round.

ZoliQua (changes requested):
I said a screenshot would settle the RTL layout better than either of us could by reading, so I went and got one. I could not run WeasyPrint, but the PDF is built from HTML, so I rendered `_generate_html(...)` for `ar` and `es` with Arabic content and opened both in a browser.

**Good news first:** the mirroring works. Clinic block top-right, invoice number and dates top-left, table columns in reverse order (`#` rightmost), totals block right-aligned, Arabic text shapes correctly, and money renders as `60.00 €` with the Arabic decimal convention. Nothing is broken.

**One real defect, and it is one line.** In the line-items table, the `Description` header sits at the opposite edge of the page from the descriptions it labels. Measured in the DOM on the rendered page:

| | `text-align` (computed) | column left | column width |
|---|---|---|---|
| `ar` — `<th>` الوصف | `left` | 446 | 1177 |
| `ar` — `<td>` تنظيف الأسنان | `start` | 446 | 1177 |

Same column, 1177 px wide: the header resolves `left` physically, the cell resolves `start` logically (which is *right* under `dir="rtl"`), so the label ends up about 1,100 px away from its own data — it reads as a stray word floating in the middle of the table. In the `es` render the same two elements both land at `left: 106`, which is why nobody has seen it before: `left` and `start` are the same edge in LTR, and this PR is what first sets `dir="rtl"`.

The fix is `pdf.py:399`:

```css
th {
    text-align: start;   /* was: left */
}
```

Same class of issue, but *not* currently visible, so only if you want to be thorough: `.price`, `.discount` and `.total` (`pdf.py:416-419`) use physical `right`. Those columns are 80–100 px wide, so the mirrored result still looks fine; `end` would be the logically correct keyword.

I am switching to request-changes only because it is a one-liner in the PR that introduces RTL — everything else about the change is ready, and I will re-approve as soon as it is in. One `dir="rtl"` heading assertion cannot catch this; if you want it pinned, asserting that `th` and `td` in the same column resolve to the same side is hard in a unit test, so I would let the fix stand on its own.

Method, if you want to reproduce: build the HTML with `InvoicePDFService._generate_html(invoice, clinic, is_preview=False, locale="ar")`, write it to a file, serve the folder with `python3 -m http.server`, and open it. No WeasyPrint needed — it only turns that same HTML into a PDF.

ZoliQua:
Filed #485 for something the same render turned up — not for this PR, just so the thread connects.

Rendering the **budget** PDF in Arabic the same way shows it is still almost entirely English: `Quote`, `Patient Information`, `Treatments`, `Description`, `Qty`, `Unit Price`, `Subtotal`, `TOTAL:`, `Patient Signature`. Only the clinic name and treatment names are Arabic, because those come from the data. `budget/pdf.py:723` is `return labels_es if locale == "es" else labels_en`, while the route accepts all ten `PDF_LOCALE_PATTERN` locales — the same dead-end #422 described for invoices, one module over. Purchase orders are in the same state and additionally pick their money locale with `"es_ES" if locale == "es" else "en_US"`.

It also has the two RTL items from my review above: no `dir="rtl"` for `ar`, and the same `th { text-align: left }` rule.

If you want to take #485 after this lands, the invoice work here is the template to copy — that is most of why I wrote it up with the mapping table.

lamanji:
@ZoliQua done in `d4a0579e` - `th` is now `text-align: start`, and I took the thorough option too: `.price`, `.discount` and `.total` are `end` instead of `right`.

No test, per your instruction - the fix stands on its own. pdf-format suite 8 passed, ruff clean.

ZoliQua (approved):
Approving — re-rendered the Arabic invoice from `d4a0579e` and checked the DOM rather than trusting the diff:

```
descHeader: { align: "start", left: 446, width: 1177 }
descCell:   { align: "start", left: 446, width: 1177 }
```

Header and cell now resolve to the same edge of the same column, so `الوصف` sits above its own data instead of across the page. The amount columns are `end` as well — thank you for taking the thorough option; they line up cleanly against the mirrored table.

Agreed on no test for it: asserting that two elements resolve to the same side is awkward to express and would pin the markup rather than the behaviour.

Two physical rules are left in that file, both cosmetic and neither worth another round: `.invoice-info` is still `text-align: right` (measured `align: "right"`, so in Arabic the header block's ragged edge faces the page centre instead of the outer edge, where its LTR twin puts it), and `.totals .label` / `.value` are still `left`/`right`. The totals read correctly as rendered because the table itself mirrors. If you touch this file again, `start`/`end` would finish the job — and #487 has the same two in budget and purchase orders, so I will take them there rather than send you round again.

